PR #376, opened 2026-03-14
Refactor runtime path and env resolution

basnijholt:
## Summary
- centralize mutable runtime config, storage, and env authority around `RuntimePaths` with a pure `resolve_runtime_paths()` path and an explicit `activate_runtime_paths()` path
- replace stale import-time runtime snapshots with runtime-aware accessors across runtime consumers, and align validation/runtime behavior for config-relative and agent-owned path domains
- add regression guardrails for placeholder rejection, cross-workspace isolation, runtime-global/env allowlists, and modules imported before runtime activation changes

## Runtime contract
- `RuntimePaths` is now the single source of truth for active config path, config dir, config-adjacent `.env`, shared storage root, and runtime-varying env-derived settings
- env precedence is now: explicit runtime args, true exported shell env, sibling `.env`, then code defaults
- config-relative paths only allow `MINDROOM_CONFIG_PATH` and `MINDROOM_STORAGE_PATH` placeholders
- agent-owned paths remain workspace-relative and reject absolute paths and env-var placeholders
- worker runtime roots stay backend-controlled and separate from canonical durable agent state

## Cleanup
- remove duplicated/local runtime path resolution from CLI config init, knowledge manager path resolution, cache/logging/storage consumers, and sandbox fallback storage handling
- remove stale import-time consumers of mutable runtime globals like Matrix SSL/server-name/namespace and AI cache settings
- regenerate `skills/mindroom-docs` references after CLI help formatting updates from hooks

## Verification
- `pre-commit run --all-files`
- `uv run pytest -q`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -109,7 +109,13 @@ def orchestrator_runtime_paths(
     *,
     config_path: Path | None = None,
 ) -> RuntimePaths:
-    """Build an explicit runtime context for orchestrator tests."""
+    """Build an explicit runtime context for orchestrator tests.
+
+    Default the config path to an isolated file under the provided test root so
+    callers never fall back to the tracked repo-root config.yaml.
+    """
+    if config_path is None:
+        config_path = storage_path / "config.yaml"
     return resolve_runtime_paths(
         config_path=config_path,
         storage_path=storage_path,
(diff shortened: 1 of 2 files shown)


src/mindroom/agents.py
```diff
@@ -691,14 +690,17 @@ def create_agent(  # noqa: PLR0915, C901, PLR0912
     """
     from mindroom.ai import get_model_instance  # noqa: PLC0415
 
-    resolved_runtime_paths = runtime_paths or constants.get_runtime_paths()
+    resolved_runtime_paths = runtime_paths or config.runtime_paths
```
basnijholt: Why do we have 2 oprions? Still seems like a smell

PR #554, opened 2026-04-11
Add thread IDs to structured log lines

basnijholt:
## Summary
- bind `thread_id` into the structured logging context across message, reaction, scheduling, delivery, and dispatch paths
- include `thread_id: null` explicitly for room-level logs so log consumers always get a stable field
- restack the message-entry logging hunk so the extracted patch is valid on current `main`

## Verification
- `python -m pre_commit run --files src/mindroom/bot.py src/mindroom/delivery_gateway.py src/mindroom/dispatch_planner.py src/mindroom/interactive.py src/mindroom/logging_config.py src/mindroom/response_coordinator.py src/mindroom/scheduling.py tests/test_logging_config.py tests/test_bot_scheduling.py`
- `PYTHONPATH="$PWD/src" .venv/bin/pytest -q tests/test_logging_config.py tests/test_bot_scheduling.py -k "thread_id or logging"`

### DIFF
--- a/src/mindroom/bot.py
+++ b/src/mindroom/bot.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Literal, cast
 from uuid import uuid4
 
 import nio
+from structlog.contextvars import bind_contextvars, bound_contextvars
 from tenacity import retry, retry_if_not_exception_type, stop_after_attempt, wait_exponential
 
 from mindroom.bot_runtime_view import BotRuntimeState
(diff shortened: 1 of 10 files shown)


basnijholt:
Closing this in favor of the newer `MessageTarget`-based logging work in #556, which merged on April 11, 2026.

Why not merge/rebase this PR:
- the core abstraction change is already in `main`
- this branch now overlaps heavily with merged code
- the remaining diff is a mix of old plumbing and extra logging coverage, so rebasing it would be noisier than the value it adds

What to do instead:
- start a fresh follow-up PR from current `main`
- keep only the remaining unique logging coverage that is still missing after #556
- focus that follow-up on the leftover ingress/background paths such as `bot.py`, `scheduling.py`, `stop.py`, and any still-uncovered delivery logs
- keep it small and verify it against current `main`, rather than reviving the old pre-`MessageTarget` diff

PR #557, opened 2026-04-11
Fix remaining thread log context gaps

basnijholt:
## Summary
- follow up to #556 with the remaining current-main thread log context fixes called out from the #554 discussion
- merge bound contextvars into the foreign logger JSON processor chain and bind scheduled workflow execution to the target room/thread context
- include the inbound bot receive log in thread context coverage and add regression tests for bot, scheduler, foreign logger, delivery, and stop-manager logging paths

## Test Plan
- uv run pytest -n 0 -q

### DIFF
--- a/src/mindroom/bot.py
+++ b/src/mindroom/bot.py
@@ -1578,14 +1578,20 @@ class AgentBot:
 
     async def _handle_message_inner(self, room: nio.MatrixRoom, event: nio.RoomMessageText) -> None:
         """Handle one text message inside the per-turn thread-history cache scope."""
-        self.logger.info("Received message", event_id=event.event_id, room_id=room.room_id, sender=event.sender)
+        event_info = EventInfo.from_event(event.source)
+        self.logger.info(
+            "Received message",
+            event_id=event.event_id,
+            room_id=room.room_id,
+            sender=event.sender,
+            thread_id=event_info.thread_id or event_info.thread_id_from_edit,
+        )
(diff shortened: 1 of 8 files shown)


basnijholt:
I think there is still one ingress gap in `src/mindroom/bot.py`.

The new `Received message` log uses `event_info.thread_id or event_info.thread_id_from_edit`, but that is only the raw Matrix relation, not the canonical MindRoom conversation target.

Later in the same flow, MindRoom resolves plain reply-chain messages through `coalescing_thread_id()` / `build_message_target(...)`, and those can still land in an existing thread even when the raw event has no native `m.thread` metadata.

So this log line can still emit `thread_id = null` for messages that the rest of the pipeline treats as threaded.

I think this should log the same canonical thread scope the pipeline actually uses, either by:
- logging after `coalescing_thread_id()` is resolved, or
- computing the target first and logging `target.resolved_thread_id`

Otherwise this PR improves ingress logging, but does not fully close the remaining bot-side thread-context gap.

basnijholt:
Fixed in `5ef44c6d`.

`Received message` now logs the canonical ingress thread scope from `coalescing_thread_id()` instead of only the raw Matrix relation fields, so plain replies that MindRoom resolves into an existing thread no longer emit `thread_id = null`.

Added regression coverage in `tests/test_bot_scheduling.py` for the plain-reply command path as well as the existing native `m.thread` path.

PR #571, opened 2026-04-12
fix: skip command parsing for voice messages (ISSUE-137)

basnijholt:
## Summary
- Rebuilt from current local main as a clean origin/main-based branch.

## Test Plan
- Not rerun during branch refresh; source changes were cut from the current local main stack.


basnijholt:
Consolidated into #572 so ISSUE-137 stays in a single PR.

PR #590, opened 2026-04-15
Ignore fallback thread metadata in Matrix DMs

basnijholt:
## Summary
- ignore fallback `m.thread` metadata on inbound Matrix DM follow-up messages
- keep DM follow-up routing in room scope instead of treating fallback metadata as a real thread
- add a regression test covering DM follow-ups that include fallback thread metadata

## Testing
- `.venv/bin/pytest -q tests/test_routing_regression.py -k 'dm_fallback_thread_message_is_routed_in_room_mode'`


basnijholt:
Closing this because the intended behavior is to keep private DMs thread-based. This change incorrectly treated DM follow-ups as room-scoped instead of preserving threaded DM conversations.

PR #619, opened 2026-04-16
fix: split room cache barriers by thread

basnijholt:
## Summary
- add thread-scoped cache coordination so thread reads and writes wait on room-wide work and same-thread work, but not unrelated threads in the same room
- route outbound and live thread mutations through the thread-specific barrier whenever the thread id is known
- add regression coverage for same-thread serialization and cross-thread non-blocking behavior in the shared event cache coordinator

## Test Plan
- [x] uv run pytest


basnijholt:
PR #619 is superseded by merged PR #716: d09b97a6c Fix thread cache barrier ordering (#716).

PR #634, opened 2026-04-17
perf: skip redundant router ingress work

basnijholt:
## Summary
- cherry-pick `19c081012` from `gitea/main` onto the current `origin/main`
- skip redundant router ingress work when the router can already determine the turn should not go through shared ingress setup
- add focused scheduling coverage for the router early-bail paths

## Verification
- uv run pytest tests/test_bot_scheduling.py -x --no-cov -q -n auto
- uv run pytest -n auto


basnijholt:
Superseded by split PRs #635 (timing helpers) and the stacked router follow-up from that branch.

PR #701, opened 2026-04-23
refactor: simplify streaming delivery orchestration

basnijholt:
## Invariant
`streaming.py` orchestrates, `streaming_delivery.py` owns async delivery sequencing, and `streaming_warmup.py` owns pure warmup rendering.

## What changed
- regrouped streaming runtime setup and shutdown into explicit orchestration helpers
- introduced typed runtime/cleanup state so `send_streaming_response()` reads as coordinator flow instead of inline cleanup policy
- kept delivery ownership and warmup rendering contracts intact
- added regression coverage for timeout cleanup and cancellation during shutdown so unfinished task handles remain retryable

## Verification
- `uv run pytest tests/test_streaming_behavior.py tests/test_streaming_e2e.py tests/test_worker_progress_routing.py -x -n 0 --no-cov`
- `uv run pytest -x -nauto --no-cov`

## Review note
The `compress_tool_results` default regression called out in review is already fixed on `origin/main` and is not part of this PR diff.

### DIFF
--- a/src/mindroom/streaming.py
+++ b/src/mindroom/streaming.py
@@ -622,13 +622,15 @@ async def _shutdown_streaming_runtime(
     if runtime.progress_task is not None:
         progress_task = runtime.progress_task
         progress_error = await _shutdown_worker_progress_drain(pump, progress_task)
-        runtime.progress_task = None
+        if progress_task.done():
+            runtime.progress_task = None
 
     delivery_error = None
     if runtime.delivery_task is not None:
         delivery_task = runtime.delivery_task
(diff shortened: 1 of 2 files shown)


basnijholt:
Closing this refactor instead of patching it further.

It made `send_streaming_response()` look cleaner, but it did not make the streaming shutdown/error lifecycle easier to reason about, and it introduced multiple subtle regressions in an already fragile failure path. That is not a good trade for a cleanup PR.

If we revisit this area, it should be a smaller PR with a stricter cleanup contract or a real ownership move into `streaming_delivery.py`, not more helper extraction inside `streaming.py`.

PR #719, opened 2026-04-24
fix(history): simplify compaction lifecycle

basnijholt:
## Summary
- use one compaction model: pre-reply compaction only when the current answer requires it, and post-reply compaction immediately after a completed visible reply when the updated session would require compaction next time
- send every compaction through the same Matrix lifecycle notice, then edit that notice with success or failure details
- remove the stale compaction `notify` config and old one-shot late notice path
- consolidate the Matrix lifecycle adapter and remove stale pass-through compaction policy wrappers

## Test Plan
- `nix-shell -I nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixos shell.nix --run 'uv run pytest tests/test_agno_history.py tests/test_compaction.py tests/test_compact_context.py tests/test_multi_agent_bot.py::TestAgentBot::test_generate_response_keeps_first_turn_follow_up_effects_in_new_thread -x -n 0 --no-cov -q'` (`142 passed`)
- `uv run pytest tests/test_compaction.py::test_post_response_compaction_uses_matrix_lifecycle_adapter tests/test_compaction.py::test_compaction_policy_classifies_trigger_and_required_modes tests/test_agno_history.py::test_classify_compaction_decision_forced_compaction_takes_priority tests/test_agno_history.py::test_classify_compaction_decision_uses_post_response_when_over_trigger_but_fits -q -n 0 --no-cov`
- `uv run pytest tests/test_compaction.py::test_post_response_effects_skip_compaction_after_non_streaming_run_failure tests/test_compaction.py::test_post_response_effects_skip_compaction_after_streaming_run_failure tests/test_multi_agent_bot.py::TestAgentBot::test_generate_response_marks_non_streaming_model_error_unsuccessful_for_post_effects tests/test_multi_agent_bot.py::TestAgentBot::test_generate_response_marks_streaming_model_error_unsuccessful_for_post_effects -q -n 0 --no-cov` (`4 passed`)
- `uv run ruff check src/mindroom/post_response_effects.py src/mindroom/response_runner.py tests/test_compaction.py tests/test_multi_agent_bot.py`
- `uv run ty check src/mindroom/post_response_effects.py src/mindroom/response_runner.py tests/test_compaction.py tests/test_multi_agent_bot.py`
- Earlier full-suite check on this branch printed `5207 passed, 28 skipped, 49 warnings in 420.91s`; the process then hung after the summary and was stopped with SIGTERM


basnijholt:
High-level review note on complexity:

I agree this is still too hard to reason about. The user-facing rule should be simpler than the current implementation model.

How the branch works right now:

1. Before a reply, history preparation classifies the pre-run session as `none`, `opportunistic`, or `required`.
2. `required` compaction runs before the model call and uses a Matrix lifecycle notice: send `Compacting history...`, then edit the same notice on success or failure.
3. `opportunistic` compaction does not run before the reply. It records a request, then post-response effects start an async task after final response delivery.
4. The old one-shot late compaction notice sender has been removed.

The problem is that this gives us two policy paths and two execution moments: pre-reply required compaction and post-reply opportunistic compaction. That makes the behavior harder to explain, and it also creates edge cases. In particular, the post-reply request is computed from the pre-reply session, so if the reply that just finished is what pushes the session over the threshold, this branch can still wait until the next turn before compacting.

The simpler model I think we should aim for:

1. There is one destructive compaction runner.
2. Every destructive compaction uses the same Matrix lifecycle notice: send once, edit once, no separate late-notice path.
3. Before the model call, run compaction only if the current answer cannot preserve acceptable history without compacting first. This is the blocking/manual path.
4. After a successful visible reply, re-open the persisted session and run the same threshold check against the updated history. If the next turn would compact, compact immediately then, using the same lifecycle notice.
5. The async task can remain as an implementation detail so the already-delivered reply is not held open, but it should not be a separate user-facing/background policy with different semantics.

That would make the docs straightforward: MindRoom compacts either before a reply when needed for that reply, or immediately after a reply when the completed turn made the next reply need compaction. In both cases the room shows a single lifecycle notice that is edited in place.

I would also clean up the config story while touching this. `notify` currently says it controls Matrix compaction notices, but this lifecycle design wants notices to be the source of truth for visible compaction. Either `notify` should still gate lifecycle notices, or it should be removed/renamed and the docs updated so users are not given a stale knob.

basnijholt:
High-level model after this simplification:

- Before an answer, history preparation compacts only when the current answer really needs durable summary context to avoid dropping material history.
- After a completed visible answer, post-response effects re-open the updated persisted session and ask the same policy question again: would the next answer require compaction now? If yes, it compacts immediately.
- Both pre-reply and post-reply compaction use the same Matrix lifecycle notice: send `Compacting history...`, then edit that same notice with success/failure stats.
- The old `notify` option and late one-shot compaction notice are gone. There should no longer be a second user-facing path for destructive history compaction.

This still uses an in-process background task only to avoid blocking after the reply has already been delivered. It is not a separate durable queue or deferred next-turn system; the work starts immediately after final delivery.

PR #783, opened 2026-04-28
Recover Postgres event cache after transient connection loss

basnijholt:
## Summary
A single transient PostgreSQL connection failure (admin shutdown, server-closed socket, psycopg `InterfaceError`) currently cascades into a permanent `disable()` of the in-process event cache for the rest of the runtime. The DB still has valid rows, but every subsequent read returns `no_cache_state`, forcing the caller to fall back to a homeserver paginate-and-scan loop.

This change:
- classifies a tight set of connection-level errors as transient (`08xxx`, `57P01-57P03`, `psycopg.InterfaceError`, plus a small allow-list of message fragments like "the connection is closed" / "terminating connection"),
- drops the dead connection on the runtime so `acquire_db_operation` reopens it,
- retries the failing read or write **once** on a fresh connection,
- leaves the permanent `disable()` path intact for non-recoverable fail-closed scenarios (e.g. `_fail_closed_thread_invalidation` after the retry has already been spent).

It also exposes `disabled_reason`, `unavailable_reason`, and `reconnect_attempts` on the cache facade so future diagnostics can explain why writes were skipped instead of just logging `cache_runtime_available=False`.

## Tests
- `uv run pytest tests/test_postgres_event_cache_recovery.py -x -n 0 --no-cov -q` → 6 passed
  - 4 unit tests covering `_is_transient_db_error` classification (sqlstate, `InterfaceError`, message-match, rejected codes)
  - 2 live-Postgres scenarios that force-close the underlying connection mid-flight and verify the next read/write reconnects, completes, and leaves `disabled_reason` unset (`reconnect_attempts == 1`)
- `uv run pytest tests/test_event_cache_backends.py tests/test_event_cache.py -nauto --no-cov -q` → 68 passed (no regression)
- `uv run ruff check src/mindroom/matrix/cache/postgres_event_cache.py tests/test_postgres_event_cache_recovery.py`

## Test plan
- [x] transient sqlstates and `InterfaceError` classified correctly
- [x] non-transient errors propagate without retry
- [x] `invalidate_thread` recovers after a force-closed connection
- [x] `get_thread_events` recovers after a force-closed connection
- [x] `disabled_reason` / `unavailable_reason` / `reconnect_attempts` exposed on the cache

### DIFF
--- a/src/mindroom/matrix/cache/postgres_event_cache.py
+++ b/src/mindroom/matrix/cache/postgres_event_cache.py
@@ -38,15 +38,15 @@ T = TypeVar("T")
 # fresh connection.
 _TRANSIENT_SQLSTATES: frozenset[str] = frozenset(
     {
-        "08000",
-        "08001",
-        "08003",
-        "08004",
-        "08006",
-        "08007",
-        "57P01",
-        "57P02",
-        "57P03",
+        "08000",  # connection_exception (generic 08xxx class)
+        "08001",  # sqlclient_unable_to_establish_sqlconnection
+        "08003",  # connection_does_not_exist
+        "08004",  # sqlserver_rejected_establishment_of_sqlconnection
+        "08006",  # connection_failure
+        "08007",  # transaction_resolution_unknown
+        "57P01",  # admin_shutdown (terminating connection due to administrator command)
+        "57P02",  # crash_shutdown (postmaster has crashed)
+        "57P03",  # cannot_connect_now (server still starting up)
     },
 )
 _TRANSIENT_MESSAGE_FRAGMENTS: tuple[str, ...] = (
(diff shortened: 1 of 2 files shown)


basnijholt:
Closing in favour of #784, which is a superset of this PR.

#784 keeps the same classifier + retry shape but additionally:
- buffers pending stale markers and replays them when the backend recovers (this PR drops them on the floor — correctness regression, not just a perf miss)
- raises a typed `EventCacheBackendUnavailable` so `_fail_closed_thread/room_invalidation` can distinguish transient backend loss from logical errors and skip the permanent `disable()`
- proactively checks `connection.closed` instead of waiting for the next op to fail
- wires the new diagnostics (`cache_postgres_disabled_reason`, `transient_failure_count`, `reconnect_count`, `pending_*_invalidations`) into the existing `matrix_sync_certification_uncertain` log line

Going with #784.

PR #840, opened 2026-05-01
Fix stored OAuth client config boundaries

basnijholt:
## Summary
- split provider-specific and shared OAuth client config so shared Google config never supplies the wrong redirect URI
- keep OAuth client config in the primary runtime store, reject copy-from/to client config services, and preserve redacted client secrets on save
- expose provider client-config services in OAuth status and add a dashboard path to configure missing OAuth clients

## Verification
- uv run pytest tests/api/test_oauth_api.py tests/api/test_credentials_api.py tests/test_credentials.py tests/test_config_manager_consolidated.py tests/test_credential_policy.py -x -n 0 --no-cov -v
- bun run test:unit src/components/Integrations/Integrations.test.tsx src/components/Integrations/integrations/index.test.ts
- uv run tach check --dependencies --interfaces


basnijholt:
Superseded by #842 after #839 was reverted. The new PR reapplies the full feature from the reverted main baseline.

PR #1032, opened 2026-05-13
Fix OpenRouter structured output fallback

basnijholt:
## Summary
- default OpenRouter models to prompt-only structured outputs instead of provider JSON mode
- keep native structured output as an explicit model option
- add regression coverage using the current OpenRouter preset

## Tests
- uv run pytest tests/test_extra_kwargs.py::test_openrouter_uses_prompt_only_structured_outputs_by_default -q -n 0 --no-cov
- uv run ruff check src/mindroom/model_loading.py tests/test_extra_kwargs.py
- uv run pytest tests/test_extra_kwargs.py tests/test_thread_summary.py tests/test_model_defaults.py -q -n 0 --no-cov
- uv run pytest -q
- uv run pre-commit run --files src/mindroom/model_loading.py tests/test_extra_kwargs.py

Note: full pytest passed with coverage warnings from the local .coverage database, but exit code was clean.

## Summary by Sourcery

Default OpenRouter models to use prompt-only structured outputs while keeping native structured output support explicitly configurable.

New Features:
- Introduce an OpenRouter wrapper model that disables native structured outputs and JSON schema outputs by default and strips response_format from requests when unsupported.

Tests:
- Add regression test ensuring the OpenRouter preset uses prompt-only structured outputs by default and does not send response_format in request parameters.

## Release Notes

* **Bug Fixes**
  * OpenRouter models now default to prompt-based structured outputs for improved API compatibility and reliability.
  * Request parameters are automatically adjusted to exclude unsupported response format specifications when native structured outputs are unavailable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mindroom-ai/mindroom/pull/1032)

### DIFF
--- a/src/mindroom/model_loading.py
+++ b/src/mindroom/model_loading.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, cast
 
 from agno.models.anthropic import Claude
@@ -27,6 +28,8 @@ from mindroom.vertex_claude_prompt_cache import install_vertex_claude_prompt_cac
 
 if TYPE_CHECKING:
     from agno.models.base import Model
+    from agno.run.agent import RunOutput
+    from pydantic import BaseModel
 
     from mindroom.config.main import Config
     from mindroom.config.models import ModelConfig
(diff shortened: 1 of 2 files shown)


basnijholt:
Closing this PR rather than merging it.

The original voice-message failure was addressed separately in #1035 by retrying the existing no-inline-media fallback path for OpenRouter's "No endpoints found that support input audio" error.

This PR targets the separate OpenRouter structured-output/JSON-mode error, but the current approach is too broad: it disables structured-output behavior for OpenRouter as a provider, even though OpenRouter routes to multiple underlying models/endpoints with different capability profiles. It also leaves the explicit opt-in path incomplete, since  can still be stripped even when JSON schema support is enabled.

A future fix should be capability-aware or explicitly configurable, not a blanket OpenRouter wrapper.

basnijholt:
Clarification on the close reason:

The original voice-message failure was addressed separately in #1035 by retrying the existing no-inline-media fallback path for OpenRouter's "No endpoints found that support input audio" error.

This PR targets the separate OpenRouter structured-output/JSON-mode error, but the current approach is too broad: it disables structured-output behavior for OpenRouter as a provider, even though OpenRouter routes to multiple underlying models/endpoints with different capability profiles. It also leaves the explicit opt-in path incomplete, since `response_format` can still be stripped even when JSON schema support is enabled.

A future fix should be capability-aware or explicitly configurable, not a blanket OpenRouter wrapper.

PR #1072, opened 2026-05-20
Fix config generation after API saves

basnijholt:
## Summary
- Prevent identical successful config reloads from advancing the API config generation.
- Add a regression for API save, watcher-style reload, then another save with the returned generation.

## Test Plan
- uv run pytest tests/api/test_api.py::test_config_generation_headers_protect_full_and_raw_save_endpoints tests/api/test_api.py::test_config_reload_after_api_save_keeps_returned_generation tests/api/test_api.py::test_first_party_config_writers_advance_generation_before_watcher_reload tests/api/test_api.py::test_load_config_into_app_discards_stale_results_after_runtime_swap tests/api/test_api.py::test_load_config_into_app_ignores_runtime_mismatches_after_api_runtime_swap -q
- git diff --check
- commit hooks

## Summary by Sourcery

Adjust config reload behavior to avoid advancing the config generation on identical successful reloads and add regression coverage around API save and watcher-driven reload interactions.

Bug Fixes:
- Prevent config generation from advancing when a watcher-triggered reload successfully reapplies an unchanged configuration, so API save responses are not made immediately stale.

Tests:
- Add regression test ensuring config reload after an API save preserves the returned generation and allows a subsequent save using that generation.

### DIFF
--- a/src/mindroom/api/config_lifecycle.py
+++ b/src/mindroom/api/config_lifecycle.py
@@ -135,19 +135,18 @@ def _source_fingerprint(source: bytes | str) -> str:
     return hashlib.sha256(source_bytes).hexdigest()
 
 
-def _config_source_fingerprint(runtime_paths: constants.RuntimePaths) -> str:
-    """Return the raw source identity for the active config file."""
-    return _source_fingerprint(runtime_paths.config_path.read_bytes())
-
-
 def _load_config_result(
     runtime_paths: constants.RuntimePaths,
 ) -> tuple[ConfigLoadResult, dict[str, Any] | None, Config | None, str | None]:
     """Load and validate one config file without mutating shared app state."""
     source_fingerprint: str | None = None
     try:
-        source_fingerprint = _config_source_fingerprint(runtime_paths)
-        runtime_config = load_runtime_config_model(
+        source_bytes = runtime_paths.config_path.read_bytes()
+        source_fingerprint = _source_fingerprint(source_bytes)
+        source_text = source_bytes.decode("utf-8")
+        data = yaml.safe_load(source_text) or {}
+        runtime_config = Config.validate_with_runtime(
+            data,
             runtime_paths,
             tolerate_plugin_load_errors=True,
         )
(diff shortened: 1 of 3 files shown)


basnijholt:
Addressed in cc7e7bf64: the API snapshot now tracks the raw config source fingerprint. Watcher reloads over the exact source already published by an API save keep the returned generation, while external raw-only edits such as comments/formatting advance generation and make stale /api/config/raw saves return 409. Added regression coverage for both flows and asserted the follow-up save advances generation.

basnijholt:
Addressed the latest config reload race in 053d9fbc8. The API reload path now reads the raw config bytes once, fingerprints those exact bytes, and validates that same source, so a generation maps to one concrete source snapshot rather than a second read of the same path.\n\nFocused checks run:\n- uv run pytest tests/api/test_api.py::test_config_reload_uses_source_fingerprint_from_validated_source -q -n 0 --no-cov\n- uv run pytest tests/api/test_api.py::test_config_generation_headers_protect_full_and_raw_save_endpoints tests/api/test_api.py::test_config_reload_after_api_save_keeps_returned_generation tests/api/test_api.py::test_external_raw_config_reload_advances_generation_for_same_authored_config tests/api/test_api.py::test_config_reload_uses_source_fingerprint_from_validated_source tests/api/test_api.py::test_first_party_config_writers_advance_generation_before_watcher_reload tests/api/test_api.py::test_load_config_into_app_discards_stale_results_after_runtime_swap tests/api/test_api.py::test_load_config_into_app_ignores_runtime_mismatches_after_api_runtime_swap tests/api/test_api.py::test_config_lifecycle_published_snapshot_owns_optional_runtime_fields tests/api/test_api.py::test_initialize_api_app_clears_config_cache_when_config_path_changes tests/api/test_api.py::test_initialize_api_app_clears_config_cache_when_runtime_changes tests/api/test_api.py::test_api_config_load_does_not_serve_stale_cache_after_invalid_reload tests/api/test_api.py::test_api_cached_write_endpoints_refuse_stale_config_after_invalid_reload -q -n 0 --no-cov\n- uv run ruff check src/mindroom/api/config_lifecycle.py tests/api/test_api.py\n- uv run ruff format --check src/mindroom/api/config_lifecycle.py tests/api/test_api.py\n- git diff --check -- src/mindroom/api/config_lifecycle.py tests/api/test_api.py

PR #1073, opened 2026-05-20
Restart sync loop after unexpected clean return

basnijholt:
## Summary
- Treat a clean Matrix sync loop return as intentional only after the bot has stopped
- Restart the sync loop when it returns while the bot is still marked running
- Add regression coverage for the stale running-but-not-syncing state

## Tests
- uv run pytest tests/test_sync_task_cancellation.py -q
- uv run ruff check src/mindroom/orchestration/runtime.py tests/test_sync_task_cancellation.py
- pre-commit hooks during commit

## Summary by Sourcery

Handle unexpected clean sync loop returns while the bot is still marked as running and add regression coverage for this scenario.

Bug Fixes:
- Prevent bots from remaining present but permanently stopping sync when the sync loop returns cleanly while still marked running.

Tests:
- Add regression test ensuring the sync loop restarts when sync_forever returns cleanly while the bot is still marked running.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sync restart behavior to correctly handle cases where the sync operation returns while the bot is still running.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mindroom-ai/mindroom/pull/1073?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/mindroom/orchestration/runtime.py
+++ b/src/mindroom/orchestration/runtime.py
@@ -534,7 +534,15 @@ async def sync_forever_with_restart(bot: AgentBot | TeamBot, max_retries: int = 
                 await bot.prepare_for_sync_shutdown()
                 await iteration.cancel()
 
-        if not bot.running or (max_retries >= 0 and retry_count >= max_retries):
+        if not bot.running:
+            break
+        if max_retries >= 0 and retry_count >= max_retries:
+            logger.error(
+                "sync_loop_retries_exhausted",
+                agent=bot.agent_name,
+                retry_count=retry_count,
+                max_retries=max_retries,
+            )
             break
 
         wait_time = retry_delay_seconds(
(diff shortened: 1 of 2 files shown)


basnijholt:
Addressed the actionable Sourcery feedback in 15f4d5178.

- Verified the clean-return path already uses the existing retry/backoff path before restarting, and added a regression assertion for that.
- Added an explicit `sync_loop_retries_exhausted` error log when finite `max_retries` is exhausted while `bot.running` is still true.
- Kept stop semantics unchanged; the supervisor now reports the stale-running risk instead of forcing lifecycle state from this helper.

Local verification:
- `uv run pytest tests/test_sync_task_cancellation.py -x -n 0 --no-cov -v`
- `uv run pytest -n 0 --no-cov`

PR #1004, opened 2026-05-10
feat: TTS replies as proper Matrix voice messages

basnijholt:
Cherry-picks `7a8b5085f` from `gitea/main` onto `origin/main`.

## Summary
- Sends TTS/audio replies as Matrix voice messages with MSC voice metadata when requested.
- Adds voice payload generation and fallback behavior for regular audio messages.
- Extends matrix message and attachment tools to pass voice-send intent.

## Verification
- `uv run --all-extras pytest tests/test_voice_message.py tests/test_send_file_message.py tests/test_matrix_message_tool.py tests/test_attachments_tool.py -q`

## Summary by CodeRabbit

* **New Features**
  * Added Matrix voice-message support for attachments via an as_voice option, sending audio as Matrix voice events with duration, waveform, and MSC metadata.
  * Automatic transcoding to Opus-in-Ogg when needed, with temporary-file creation and robust cleanup across success, failure, and cancellation paths.

* **Tests**
  * Expanded test coverage for voice preparation, transcoding, sending, fallback behavior, cleanup, cancellation, and encrypted-room delivery.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mindroom-ai/mindroom/pull/1004)

PR #1114, opened 2026-06-02
fix(history): clear Claude thinking for compaction summary calls

basnijholt:
Cherry-picks `44fdca3034d9389ab943f4021202389830f22305` from `gitea/main` as a stacked follow-up on #1009, because it depends on the compaction tuning API introduced there.\n\nSkipped `6ee8e68fe` per request. Existing open PRs already cover `08cf97152` (#1004) and `3b2fd0caa` (#1009).

## Summary by Sourcery

Clear Claude-specific thinking settings when configuring models for compaction summaries.

Bug Fixes:
- Ensure compaction-configured models reset any existing Claude thinking configuration to avoid reply-oriented behavior during summaries.

Tests:
- Extend compaction configuration tests to cover clearing of Claude thinking settings.

Chores:
- Update the vulture whitelist for the new thinking attribute on models.


basnijholt:
Superseded by #1264 (merged as 876252406). It absorbs this fix — `thinking = None` for Claude summary calls — inside `configure_summary_model` in `history/summary_call.py`, the same single function that caps `max_tokens`, since the two constraints are coupled (a thinking budget at or above the cap is a 400 from Anthropic). Covers anthropic, vertexai_claude, and bedrock_claude via the shared `Claude` base class, pinned by invariant 3 in `tests/test_compaction_invariants.py`.

PR #1134, opened 2026-06-02
Keep unavailable plugin tools out of runtime metadata

basnijholt:
## Summary
- Keep skipped broken-plugin tool metadata out of public runtime/catalog metadata.
- Preserve unavailable skipped-plugin tools only in validation snapshots so tolerant config validation can still filter and warn.
- Add a regression for a pre-registration broken plugin tool being validation-only.

## Test Plan
- [x] Focused broken-plugin metadata regression passed.
- [x] Ruff and pre-commit checks passed.
- [x] Broader plugin/tool metadata tests passed.
- [x] API/facade tool metadata slice passed.
- [ ] Full suite was started serially and interrupted by request; future full runs should use `-n auto`.

## Summary by Sourcery

Ensure unavailable plugin tools are excluded from public runtime tool metadata while remaining visible for validation snapshots.

Enhancements:
- Adjust runtime tool state resolution so tools from broken or unavailable plugins are no longer included in runtime tool metadata but are tracked separately as unavailable metadata.
- Update validation snapshot construction to merge unavailable plugin tool metadata back into validation-only state while still marking them as unavailable.

Tests:
- Add a regression test verifying that a pre-registration broken plugin tool is omitted from runtime metadata but present and correctly flagged in validation metadata.

### DIFF
--- a/src/mindroom/tool_system/metadata.py
+++ b/src/mindroom/tool_system/metadata.py
@@ -766,6 +766,15 @@ class ToolMetadata:
     factory: Callable | None = None  # Factory function to create tool instance
 
 
+@dataclass(frozen=True)
+class _ResolvedToolState:
+    """Runtime-visible tool state plus validation-only skipped-plugin metadata."""
+
+    tool_registry: dict[str, Callable[[], type[Toolkit]]]
+    tool_metadata: dict[str, ToolMetadata]
+    unavailable_tool_metadata: dict[str, ToolMetadata]
+
+
 def register_tool_with_metadata(
     *,
     name: str,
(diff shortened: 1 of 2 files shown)


basnijholt:
Addressed the tuple-clarity feedback in 2cd34b741 by replacing the positional return tuple from `_resolved_tool_state_for_runtime` with a private typed `_ResolvedToolState` dataclass and updating the call sites to use named fields.

I intentionally did not change the validation path to mutate `desired_metadata` in place. Keeping `validation_metadata` as a distinct merged mapping preserves the runtime/public metadata vs validation-only metadata separation that this PR is fixing.

PR #1142, opened 2026-06-02
Emit room-member join hooks for prev-less live joins

basnijholt:
## Summary
- emit room-member join hooks for live Matrix member events even when `unsigned.prev_content` is absent
- apply the same behavior to restored-token timeline catch-up events
- keep full-state sync snapshots guarded by previous membership so existing members are not replayed as new joins

## Tests
- `uv run pytest tests/test_room_member_hooks.py -q`

### DIFF
--- a/src/mindroom/bot.py
+++ b/src/mindroom/bot.py
@@ -93,7 +93,6 @@ from .matrix.client_room_admin import get_joined_rooms
 from .matrix.client_session import PermanentMatrixStartupError
 from .matrix.room_member_joins import (
     RoomMemberJoin,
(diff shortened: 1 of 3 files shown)


basnijholt:
Addressed the profile-update edge case from review.

Follow-up commit seeds room-member join tracking from certified sync state snapshots without emitting hooks. That keeps full-state snapshots non-triggering while preventing a later prev-less profile update for an already-known member from being treated as a first join.

Verification:
- `uv run pytest tests/test_room_member_hooks.py -q`
- pre-commit hooks during commit

PR #1240, opened 2026-06-12
Release coalescing gate at response start instead of response completion

basnijholt:
## Problem

A coalescing gate's drain awaits the entire response generation for a dispatched batch (`handle_batch` spans placeholder, streaming, tool calls, completion). Agent turns can legitimately run for minutes, and the gate holds ingress ordering the whole time:

- All of one user's top-level messages in a room share the coalescing key `(room, thread=None, requester)`, so the next top-level message queues behind the in-flight batch for the **full duration of the previous turn**. We measured a gate wait of 6.5 minutes (`coalescing_gate_flush_slow` with `oldest_pending_age_ms: 389262`) for a message whose pipeline took ~2 s once the gate released.
- Thread replies wait on the requester's root gate via `_older_owner_root_gates` while it has older queued work, so an agent writing in one thread blocks the user's messages in **other threads** of the same room.

The root cause is that the gate conflates ingress **ordering** (which must be held only briefly so messages are observed in order) with response **execution** (which can take minutes).

## Fix

A dispatched batch now carries a `ResponseStartSignal`. The response runner marks it through the existing `on_lifecycle_lock_acquired` hook at the moment it takes the per-thread response lock — i.e. once the response's target is resolved and same-thread coherence is owned by the response lock. The gate waits only until the dispatch completes **or** the response starts, whichever is first.

Started responses keep running on tracked tasks with explicit lifecycle semantics:

- unbounded drains (`drain_all`) await them before reporting completion,
- bounded drains cancel them and count the cancellation,
- a response that fails after the gate released still closes gate-owned dispatch metadata exactly once.

## Tests

New coverage in `tests/test_coalescing.py` / `tests/test_live_message_coalescing.py`:

- a second root batch dispatches once the first response has *started* (not completed),
- a thread reply dispatches while a root response is still running,
- `drain_all` awaits started dispatches,
- a bounded drain cancels started dispatches and reports incomplete,
- the failure path closes segment metadata,
- the response-start signal reaches the executor as the lifecycle-lock callback.

Full suite: 8051 passed, 48 skipped. Lint (`ruff`), `ty`, and pre-commit hooks pass.

One pre-existing test-env observation worth noting separately: rendering a large exception via `logger.exception` (structlog + rich in tests) can hold the event loop past tight polling deadlines; one new test waits on an explicit close signal instead of polling for this reason.

## Summary by CodeRabbit

* **Refactor**
  * Improved dispatch execution lifecycle management with finer-grained control over response start signaling.
  * Enhanced shutdown drain coordination to better handle concurrent dispatch task completion and cancellation.

* **Tests**
  * Added comprehensive test coverage for dispatch coalescing and drain behavior scenarios.

### DIFF
--- a/src/mindroom/coalescing.py
+++ b/src/mindroom/coalescing.py
@@ -207,6 +207,13 @@ class CoalescingGate:
     State machine per (room, thread, sender) key:
     IDLE (absent) -> DEBOUNCE -> GRACE (optional, wait for images) ->
     flush -> IN_FLIGHT, while all undispatched work remains in one FIFO queue.
+
+    The gate owns ingress ORDERING, not response EXECUTION: a dispatched batch
+    holds its gate only until the executor marks ``batch.response_start``
+    (target resolved, per-thread response lock held) or the dispatch callback
+    returns, whichever happens first. Started responses keep running on
+    gate-tracked background tasks so later batches for other targets dispatch
+    concurrently while drains can still settle or cancel them.
     """
 
     def __init__(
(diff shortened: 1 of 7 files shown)


basnijholt:
Superseded by #1254 (the #1241 redesign), which removes the full-response-duration gate hold structurally — the started-dispatch tracking this PR added is no longer needed.

PR #1249, opened 2026-06-12
Unify memory backends behind one MemoryBackend protocol

basnijholt:
## Summary

`src/mindroom/memory/` had two backends that looked like adapters but had no shared interface: parallel-but-inconsistent function sets (sync vs async), and `functions.py` branching on backend type at every call site — the seam leak behind the mixed-backend team-memory bug class.

- New `src/mindroom/memory/_backend.py` with a single `MemoryBackend` protocol (`add`, `search`, `list_all`, `get`, `update`, `delete`, `store_conversation`, `load_entrypoint_context`, `context_label`) and `resolve_memory_backend(scope, config, runtime_paths) -> MemoryBackend | None`.
- The parallel module-level backend functions fold into `FileMemoryBackend` and `Mem0MemoryBackend` adapters (no importers outside `memory/` existed; the old entry functions are deleted). The file backend's sync internals are wrapped behind the async protocol so callers no longer care.
- All `use_file_backend`-style branching in `memory/functions.py` collapses into "resolve once, call protocol methods" (~500 → ~290 lines); **public facade signatures unchanged**, so no consumers are touched.
- Now-internal team backend predicates made private in `_policy.py`; dead `agent_name=None` branches dropped. (`build_team_user_id` was already deduplicated into `_policy.py`.) `auto_flush.py` has only file-only feature gates, not backend dispatch, so it needs no protocol migration.
- `tach.toml` updated for the new module graph; `MemoryBackend`/`resolve_memory_backend` declared as the module's symbol-level interface.
- Backend selection semantics, storage layout, scope IDs, timing labels, and logs are all unchanged — this is interface unification only. Resolution lives in `_backend.py` instead of `_policy.py` to avoid an import cycle (`_file_backend`/`_mem0_backend` → `_policy`).

## Relationship to PR #798 (open — NOT absorbed)

#798 fixes mixed-backend team memory writes by partitioning team members per effective backend. This refactor deliberately does **not** change mixed-team behavior (mixed team → mem0, exactly as on main, pinned by a contract test with a comment pointing at #798). It gives that fix an obvious home: a partitioned-team `MemoryBackend` implementation composed of the two adapters and returned by `resolve_memory_backend` for mixed teams — no facade changes required. Recommend rebasing #798 onto this seam; its ~400 lines of mixed-team branching in `functions.py` become one adapter.

## Tests

- New `tests/test_memory_backend_contract.py`: one parametrized test body run against both adapters (CRUD roundtrip incl. `MemoryNotFoundError`, `list_all` scoping, cross-agent scope isolation, team vs member conversation scope, backend resolution mapping) — 11 tests.
- `uv run pytest tests/test_memory_*.py -n auto --no-cov` → 144 passed (existing suites needed only mechanical patch-target moves, 36 sites).
- Full suite → **8066 passed, 61 skipped, 0 failed**.
- `uv run tach check --dependencies --interfaces`, `uv run privata .`, `uv run pre-commit run --all-files` → pass.

Part of the 2026-06-11 architecture-review refactor wave (`refactor-prompts/10`).

### DIFF
--- a/src/mindroom/memory/_backend.py
+++ b/src/mindroom/memory/_backend.py
@@ -1,0 +1,150 @@
+"""Shared memory backend protocol and backend resolution."""
+
+from __future__ import annotations
+
+from functools import partial
+from typing import TYPE_CHECKING, Protocol
+
+from mindroom.memory.config import create_memory_instance
+
+from ._file_backend import FileMemoryBackend
+from ._mem0_backend import Mem0MemoryBackend
+from ._policy import caller_uses_disabled_memory_backend, caller_uses_file_memory_backend
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from pathlib import Path
+
+    from mindroom.config.main import Config
+    from mindroom.constants import RuntimePaths
+    from mindroom.matrix.client_visible_messages import ResolvedVisibleMessage
+    from mindroom.tool_system.worker_routing import ToolExecutionIdentity
+
+    from ._shared import MemoryResult
+
+
+class MemoryBackend(Protocol):
+    """One storage backend behind the public memory facade.
+
+    Adapters implement every operation the facade dispatches, so call sites
+    resolve a backend once and never branch on backend type again.
+    """
+
+    context_label: str
+
+    async def add(
+        self,
+        content: str,
+        agent_name: str,
+        storage_path: Path,
+        config: Config,
+        *,
+        metadata: dict | None = None,
+        execution_identity: ToolExecutionIdentity | None = None,
+    ) -> None:
+        """Add one memory for an agent scope."""
+
+    async def search(
+        self,
+        query: str,
+        agent_name: str,
+        storage_path: Path,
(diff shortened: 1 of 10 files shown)


basnijholt:
Addressed the two Greptile findings:

1. **`context_label: ClassVar[str]` vs protocol `context_label: str`** — fixed in 64038b69. `context_label` is a per-class constant, so the protocol now declares it as `ClassVar[str]`, making the structural match explicit under strict checkers.

2. **Double backend resolution in `build_memory_prompt_parts`** — declining. Resolution is two pure config predicate calls plus a frozen-dataclass/`partial` allocation (no I/O), which is noise next to the actual memory search. Threading the resolved backend into `search_agent_memories` would add a production parameter solely for that micro-optimization, and calling `backend.search` directly would duplicate the facade's `limit` default and drop the aggregate `system_prompt_assembly.memory_search` timing label. Keeping the facade as the single source of search semantics is the smaller, simpler shape.

PR #1282, opened 2026-06-12
SaaS frontend: typed API client generated from the backend's OpenAPI schema

basnijholt:
Prompt 16 of the 2026-06-11 refactor series.

## Summary

`saas-platform/platform-frontend/src/lib/api.ts` was a hand-maintained client with untyped responses (`.json()` → `any`) and per-method error-handling boilerplate; backend changes could only break the frontend at runtime. This PR types the client against the backend's own OpenAPI schema.

### Backend (`saas-platform/platform-backend/`)
- Added response models for the frontend-called routes that still returned bare dicts, describing exactly what they already return: `GdprExportResponse`, `GdprDeletionResponse`, `GdprConsentResponse`, `GdprCancelDeletionResponse` (gdpr routes) and `PricingConfigResponse`/`PricingPlanOut` (`/pricing/config`).
- Extended `AccountWithRelationsOut` with `deleted_at`, `consent_marketing`, `consent_analytics`: the `/my/account` handler returns the full accounts row, but the under-declared model made FastAPI silently strip the fields the settings page reads (the consent/pending-deletion UI was dead at runtime).
- `response_model=None` on `/matrix-oidc/authorize`: with the real slowapi limiter, the postponed `RedirectResponse` annotation is left as an unresolved ForwardRef and `app.openapi()` crashed (tests never saw this because conftest mocks the limiter).
- New `scripts/export_openapi.py` dumps the schema; `just saas-api-types` runs export + frontend type generation.

### Frontend (`saas-platform/platform-frontend/`)
- Checked-in `openapi.json` + `src/lib/api.generated.ts` generated with `openapi-typescript` (types only, no runtime client).
- `api.ts`: one typed `request<T>()` helper owns response error handling; every endpoint method is a thin typed call. Request bodies are typed against the backend's request schemas. `apiCall` is unchanged and stays exported for raw consumers (admin pages, `useSubscription`/`useUsage`).
- Components now share backend types: `useInstance`, `dashboard/instance`, and both billing pages drop their hand-rolled `Instance`/`PricingConfig` interfaces; nullable timestamps are guarded. Removed the dead "Instance Configuration" section that rendered `instance.config`, a field the instances API has never returned.
- Fixed tests that mocked fictional response shapes: InstanceCard fixtures used `account_id` instead of `id`/`subscription_id`, and the instance-lifecycle tests asserted `status`/`billing_started` fields the action routes never returned (they return `ActionResult {success, message}`).

## Drift guard

`bun run check:api-types` regenerates `src/lib/api.generated.ts` from the checked-in `openapi.json` and fails via `git diff --exit-code` if the committed file is stale. It is CI-runnable, but no workflow currently runs any frontend checks, so it is not wired into a new workflow here; the full regeneration pipeline (schema export + type generation) is documented in the frontend README and available as `just saas-api-types`.

## Coverage note: backend routes with no frontend client method

Intentionally not added to `api.ts` (unused or accessed differently):
- `admin.py` — admin pages call these via raw `apiCall` (react-admin style generic resources)
- `subscriptions.py` (`/my/subscription*`) — `useSubscription` uses raw `apiCall`
- `usage.py` (`/my/usage`) — `useUsage` uses raw `apiCall`
- `health.py` — infra probe, not called by the SPA
- `matrix_oidc.py` — browser-redirect/Synapse flows, not XHR
- `provisioner.py` (`/system/*`) — system-to-system API
- `webhooks.py` — Stripe → backend only
- individual routes `/my/account/admin-status` and `/pricing/stripe-price/{plan}/{billing_cycle}` — no current frontend caller via the typed client

## Verification

- Backend: `uv run pytest` in `saas-platform/platform-backend/` — 396 passed, 5 skipped.
- Frontend: `bun run build` passes (Next.js production build incl. type checking); `bun run test` — 17 suites, 207 tests passed.
- `bun run check:api-types` passes on a clean checkout.
- `uv run pre-commit run --all-files` passes.
- Pre-existing, unrelated: `tsc --noEmit` errors in `saas-platform/platform-frontend/tests/` (playwright files without `@playwright/test` installed) exist on `main` and are excluded from jest/build.


basnijholt:
Consolidated into #1281 — the same task ran twice in parallel. #1281 was kept for its route-anchored generated types and non-mutating drift check; this PR's improvements were ported over in 6c2638945: the test-fixture fixes (fictional API shapes), required PublicPlan fields, and the matrix_oidc root-cause comment. #1281 additionally adds a backend schema-freshness test closing the openapi.json-vs-app gap neither PR originally guarded.

PR #1347, opened 2026-07-01
Simplify compaction module surface

basnijholt:
Part 1 of a compaction-simplification series (review of `src/mindroom/history/`). This PR is purely subtractive/mechanical: no behavior changes intended.

## Changes

- **Delete `prepare_history_for_run`** (~100 lines): a production entry point whose only callers were tests, via the `prepare_history_for_run_for_test` alias in `tests/conftest.py`. The composition now lives in the test helper itself, so tests exercise the real production seams (`prepare_scope_history` + `finalize_history_preparation`) instead of a wrapper kept alive for them.
- **Delete `estimate_preparation_static_tokens`**: exported but used by nothing (not even tests).
- **Dedupe `prepare_bound_scope_history`**: the static-token estimation and resolved-inputs blocks were duplicated verbatim in both branches; hoisted above the branch.
- **Drop the redundant replay-budget fallback** in `finalize_history_preparation`: `hard_replay_budget_tokens` and `replay_budget_tokens` are resolved together in the execution plan, so the second `if history_budget is None` fallback was a no-op (comment added).
- **Drop `_resolve_replay_threshold_tokens`**: it re-implemented the `threshold_tokens is not None` check already inside `resolve_effective_compaction_threshold`.
- **Merge the two static-token estimator classes**: `AgentStaticTokenEstimator` / `TeamStaticTokenEstimator` were structurally identical; now one `StaticTokenEstimator` plus `agent_static_token_estimator` / `team_static_token_estimator` factories.
- **Pin compaction notice metadata to `version: 3`** (matching the lifecycle payloads in `delivery_gateway.py`) instead of conditionally emitting 1/2. Nothing reads the version field, and the project explicitly avoids backward compatibility.
- `tach.toml` history interface updated in the same PR.

## Testing

- Full suite: 8945 passed, 61 skipped.
- `uv run pre-commit run --all-files` clean (except `typescript-check`, which fails on this host because `tsc` is not installed — pre-existing environment gap, unrelated).
- `uv run tach check --dependencies --interfaces` clean.

Follow-up PRs (stacked): lifecycle-wrapper collapse + force-flag clearing unification, preparation-inputs resolution, and estimation deepcopy removal.


basnijholt:
Adversarial subagent review (Fable 5): **merge-safe**. All seven claims verified against before/after sources, including a structural proof that the removed budget fallback was dead code (finalize re-derives its own plan, so test-constructed plans violating the hard/soft invariant never reached it) and that the conftest helper is faithful to the deleted production function across all 41 call sites. Findings addressed in abd82f503: the test helper now imports the production `_resolve_history_scope` instead of a verbatim copy, and the two stale `version: 1` compaction fixtures were updated. The external-consumer concern is closed: a GitHub code search for `io.mindroom.compaction` across the mindroom-ai org finds hits only in this repo (mindroom-element has zero compaction references), so the version pin is externally safe.

PR #1349, opened 2026-07-01
Resolve history-preparation inputs once, trim prepare_scope_history params

basnijholt:
Part 3 of the compaction-simplification series. **Stacked on #1348** (which stacks on #1347); retarget to `main` as the earlier PRs merge.

## Changes

`prepare_scope_history` took 17 keyword parameters, eight of which (`full_prompt`, `history_settings`, `compaction_config`, `has_authored_compaction_config`, `active_model_name`, `active_context_window`, `static_prompt_tokens`, `execution_plan`) existed only to feed a two-layer resolve-if-None cascade — no production caller used most of them.

- `prepare_scope_history` now takes one required `HistoryPreparationInputs` (the former private `_ResolvedPreparationInputs`); the resolve-if-None logic lives only in `resolve_agent_preparation_inputs`, whose job it legitimately is.
- The agent path in `execution_preparation.py` resolves inputs explicitly before the call; the test helper in `conftest.py` does the same, keeping its injection surface without production pass-throughs.
- `prepare_bound_scope_history` previously resolved inputs and then unpacked them into individual parameters, which `prepare_scope_history` **re-resolved** — re-deriving the same runtime model and execution plan on every team run. It now passes the resolved value through; the double resolution is gone. (Behavior-neutral: `resolve_runtime_model` short-circuits on an explicit `active_model_name`, so the second pass always returned identical values.)
- `ResolvedHistoryExecutionPlan.authored_compaction_config` is deleted — its only reader was the removed pass-through (`authored_compaction_enabled` carries the decision-relevant bit); vulture confirmed it dead.

## Testing

- Full suite: 8945 passed, 61 skipped.
- Pre-commit clean (typescript-check skipped: `tsc` missing on this host, pre-existing).


basnijholt:
Adversarial subagent review (Fable 5): **merge-safe**, no code changes needed. The riskiest claim — that removing the team path's second input resolution is behavior-neutral — was proven field-by-field: `resolve_runtime_model` short-circuits on explicit `active_model_name` before consulting entity_name; the context-window fallback is a pure model-name-keyed lookup; every entity-sensitive fallback was disabled by explicitly passed non-None values; and the old inner `has_authored_compaction_config` (from `execution_plan.authored_compaction_config`) always equaled the outer resolved value, so `finalize_history_preparation`'s re-derived plan is unchanged. The deleted plan field has zero remaining readers (verified with a fixed-string grep after a regex grep produced a false negative). Two nits noted, both pre-existing or metrics-attribution-only, not worth changing.

PR #1381, opened 2026-07-02
Propagate ResponseTurnContext through the team envelope and prepare chain (turn-context sweep 2/4)

basnijholt:
## Turn-context sweep 2/4: propagate `ResponseTurnContext` through the team envelope and prepare chain

Stacked on #1380. Same shape as the agent-side conversion, applied to the team chain.

### What changed

- **`team_response` / `team_response_stream` take `ctx` positionally** (after `execution_identity`), deleting 7 params each: `session_id`, `run_id`, `reply_to_event_id`, `correlation_id`, `active_event_ids`, `matrix_run_metadata`, `system_enrichment_items`. The in-function derivations of `room_id`/`thread_id`/`requester_id` from `execution_identity` are gone — the caller supplies them on the context.
- **The entries refine the incoming context** via `dataclasses.replace`: `entity_label` becomes the materialized team label (`"Team (a, b)"` — exactly what the old code constructed for the driver context), and `system_enrichment_items` gets the knowledge-availability enrichment appended after member materialization. The caller's provisional `entity_label` is deliberately overwritten.
- **`prepare_materialized_team_execution` takes `ctx`**, deleting 8 params (`reply_to_event_id`, `active_event_ids`, `room_id`, `thread_id`, `requester_id`, `correlation_id`, `matrix_run_metadata`, `system_enrichment_items`). Run metadata is built from context fields. `_team_request_log_context` likewise.
- **`resolve_run_correlation_id` deleted** from `mindroom.ai` (and its tach interface): after the teams conversion it had zero callers. Its fallback chain lives on at the boundaries — callers that could pass None now mint `uuid4().hex` explicitly (openai team prepare already did).
- **Callers build the context**: `response_runner.generate_team_response_helper_locked` (one construction feeding both the streaming and blocking team calls) and `api/openai_compat._prepare_openai_team_prompt`.
- `user_id` stays a parameter on the team entries: it is the raw Agno identity for `team.arun` and the prompt `current_sender_id`, distinct from the resolved `ctx.requester_id` (`user_id or execution_identity.requester_id`).

### Behavior preservation

Intended behavior-preserving. Parity notes for review:
- Old in-entry derivations vs new caller-built fields: `execution_identity.room_id` is `target.room_id` (== `request.room_id`), `execution_identity.resolved_thread_id or .thread_id` is `target.resolved_thread_id` (both set from the same value in `build_execution_identity`), and `requester_user_id or execution_identity.requester_id` matches the old `user_id or execution_identity.requester_id` including the empty-string-falsy case.
- The old `resolve_run_correlation_id` fallback arms (metadata correlation → reply_to) were dead for every caller: response_runner always passed a resolved id; openai passed None with no reply_to/metadata (→ uuid); tests that exercised the reply_to fallback now pass that value explicitly.
- The locked-pair terminal arms in `response_runner.py` remain untouched.

Part of the stacked series 1/4 → 2/4 (this) → 3/4 execution_preparation chain → 4/4 collector conversion.

### DIFF
--- a/src/mindroom/ai.py
+++ b/src/mindroom/ai.py
@@ -1118,10 +1118,6 @@ async def _prepare_agent_and_prompt(
     )
     _mark_pipeline_timing(pipeline_timing, "memory_prepare_ready")
 
-    resolved_session_id = ctx.session_id
-    if resolved_session_id is None and scope_context is not None and scope_context.session is not None:
-        resolved_session_id = scope_context.session.session_id
-
     _mark_pipeline_timing(pipeline_timing, "agent_build_start")
 
     def _resolve_model_and_build_agent() -> tuple[ResolvedRuntimeModel, Agent]:
@@ -1140,7 +1136,7 @@ async def _prepare_agent_and_prompt(
             agent_name,
             config,
             runtime_paths,
-            session_id=resolved_session_id,
+            session_id=ctx.session_id,
             history_storage=scope_context.storage if scope_context is not None else None,
             active_model_name=runtime_model.model_name,
             knowledge=knowledge,
--- a/src/mindroom/ai.py
+++ b/src/mindroom/ai.py
@@ -116,7 +116,6 @@ __all__ = [
     "ResponseTurnContext",
     "ai_response",
     "build_matrix_run_metadata",
-    "resolve_run_correlation_id",
     "stream_agent_response",
 ]
 AIStreamChunk = str | RunContentEvent | RunCompletedEvent | ToolCallStartedEvent | ToolCallCompletedEvent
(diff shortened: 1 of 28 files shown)


basnijholt:
Adversarial review: **could not refute behavior preservation — no blockers or majors.** Derivation parity proven against build_execution_identity (room/thread/requester value-identical incl. empty-string user_id); ctx refinement ordering audited closure-by-closure; the deleted resolve_run_correlation_id fallback arms were provably dead for every production caller. The one MINOR (make_turn_context's session default silently converting formerly session-less team tests) was fixed in 3f86cd9a7. NITs disclosed: correlation contract tightened to str (production never passed None); empty source_event_id would now propagate instead of minting a uuid (unreachable today).

PR #1382, opened 2026-07-02
Propagate ResponseTurnContext through the execution-preparation chain (turn-context sweep 3/4)

basnijholt:
## Turn-context sweep 3/4: propagate `ResponseTurnContext` through the execution-preparation chain

Stacked on #1381. Completes the downward propagation: the context now flows caller → envelope → prepare chain → history preparation without any field re-exploding into kwargs along the way.

### What changed

- **`prepare_agent_execution_context` takes `ctx`**, deleting `agent_name` (→ `ctx.entity_label`), `room_id`, `thread_id`, `reply_to_event_id`, `active_event_ids`.
- **`prepare_bound_team_run_context` / `_prepare_bound_team_execution_context` take `ctx`**, deleting `room_id`, `reply_to_event_id`, `active_event_ids`. (`entity_name`/`team_name` stay — they are the configured-team key, distinct from the display `entity_label`.)
- **`_prepare_execution_context_common` takes `ctx`**, deleting `reply_to_event_id`, `active_event_ids`.
- Call sites updated: `ai._prepare_agent_and_prompt` and `teams.prepare_materialized_team_execution` now pass the context straight through instead of exploding its fields.
- The `ResponseTurnContext` import in `execution_preparation.py` is TYPE_CHECKING-only (annotation), matching the module's existing convention for `Config`/history types; tach's runtime-import graph is unchanged.

### Behavior preservation

Purely mechanical: every deleted parameter is read from the context field carrying the identical value (the callers were already passing `ctx.room_id`, `ctx.reply_to_event_id`, etc. after 1/4 and 2/4). No logic changes. The per-attempt parameters (`prompt`, `current_sender_id`, `current_timestamp_ms`, `current_prompt_is_structured`) remain explicit because dynamic-tool continuations vary them per attempt.

Part of the stacked series 1/4 → 2/4 → 3/4 (this) → 4/4 collector conversion.

## Summary by Sourcery

Propagate the ResponseTurnContext object through agent and team execution-preparation APIs and callers so execution and history preparation derive identifiers and event metadata directly from the shared turn context instead of individual parameters.

Enhancements:
- Update prepare_agent_execution_context, _prepare_bound_team_execution_context, _prepare_execution_context_common, and prepare_bound_team_run_context to accept a ResponseTurnContext instance and read room, thread, and event identifiers from it.
- Simplify ai and teams call sites to pass the turn context object directly rather than expanding its fields into keyword arguments.
- Adjust attachment and model resolution logic to use room and thread identifiers from the shared turn context, keeping behavior intact.

Tests:
- Update execution preparation, team media fallback, system enrichment, and OpenAI compatibility tests to construct and pass turn context objects, and to accommodate the new ctx parameter in fakes and helpers.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,7 +64,8 @@ Matrix sync callback
        -> ingress_lanes.py                                       (per-(room, sender) receipt-order FIFO; STT readiness waits here)
        -> coalescing.py                                          (text dispatches immediately; media-tailed batches debounce)
        -> text_ingress_dispatch.py + turn_policy.py              (ignore / route / respond decision, command execution)
-       -> response_runner.py -> ai.py                            (lifecycle lock, Agno agent/team run)
+       -> response_runner.py -> ai.py / teams.py                 (lifecycle lock, entity envelopes)
+            -> response_turn.py                                  (shared blocking/streaming turn drivers)
        -> streaming.py + delivery_gateway.py                     (progressive edits, Matrix send)
        -> turn_store.py / handled_turns.py                       (durable dedup so restarts don't double-reply)
 ```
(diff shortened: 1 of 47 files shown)


basnijholt:
Adversarial review: **behavior preservation holds — no blockers, no majors.** Field-by-field equivalence, TYPE_CHECKING-import safety (tach.toml header explicitly excludes type-only imports from depends_on), caller exhaustiveness, and merge-tree cleanliness of the session-restoration merge all verified; 515 tests across the nine affected files green. One MINOR process note: the intermediate src-only commit df267397a is not an atomic tested state (tests migrate in the follow-up commit) — moot under squash-merge, disclosed for bisect awareness.

PR #1401, opened 2026-07-03
Rewrite four apply_post_response_effects patch sites against the collaborator seam

basnijholt:
Part of the worst-offender test rewrite campaign (refactor prompt 07, deferred second half of #1393–#1398). One patch-target cluster per PR; this PR takes `mindroom.response_lifecycle.apply_post_response_effects` (31 patch sites measured on `68614c39a`; 4 rewritten here). Sibling of #1400 (independent, no stacking; both add to `ai_user_id_helpers.py`, so whichever merges second may need a trivial import-region rebase).

## What changed

The table in the refactor prompt names `PostResponseEffectsSupport` as the owning seam. Every effect inside `apply_post_response_effects` (interactive registration, run-metadata linkage, memory persistence, thread summaries) is guarded on a field of the `PostResponseEffectsDeps` that `PostResponseEffectsSupport.build_deps` constructs. So instead of patching the module function to an `AsyncMock`, tests now swap the collaborator:

- `_InertPostResponseEffects(PostResponseEffectsSupport)` — typed subclass whose `build_deps` returns `PostResponseEffectsDeps(logger=...)` only. The **real** `apply_post_response_effects` still runs each test, exercising its guard logic instead of being stubbed out.
- `_install_inert_post_response_effects(coordinator)` — rebuilds `coordinator.deps` with the inert support via `dataclasses.replace`, preserving `conversation_cache` and the other support fields the runner reads directly.

## Per-test disposition (all (a): the patch bypassed the collaborator that owns effect construction)

1. `test_generate_team_response_appends_matrix_tool_prompt_context`
2. `test_generate_team_response_allows_explicit_private_ad_hoc_member`
3. `test_generate_team_response_passes_resolved_correlation_id_to_team_response`
4. `test_generate_team_response_helper_persists_interrupted_history_when_stream_delivery_fails`

Each: drop the `apply_post_response_effects` patch line, add one `_install_inert_post_response_effects(coordinator)` call. The `should_use_streaming` / `team_response` patches in the same stacks are a different cluster (#1400's target and an accepted module-boundary patch respectively) and are deliberately untouched — one idea per PR.

## Mutate-and-check evidence

Each mutation applied to production, the test run, then reverted by inverse edit:

- Team `_append_matrix_prompt_context` call site: `include_context=include_matrix_prompt_context` → `False` — test 1 **FAILED**.
- `allow_direct_private_agents`: `and bool(execution_identity.requester_id)` → `and not bool(...)` — test 2 **FAILED**.
- `team_turn_ctx`: `correlation_id=response_identity.correlation_id` → `None` — test 3 **FAILED**.
- `interrupted_replay._render_interruption_summary`: truncated the summary string — test 4 **FAILED**.

## Counts

- Collected test count: unchanged (no tests merged, renamed, or deleted).
- `apply_post_response_effects` patch sites in `test_response_runner_team_streaming.py`: 16 → 12.
- Full suite: 9119 passed, 61 skipped. Pre-commit clean (`SKIP=typescript-check`).

## Summary by CodeRabbit

* **Tests**
  * Improved response-handling test support for team streaming scenarios, making behavior around post-response effects more consistent and reliable across coverage.

### DIFF
--- a/tests/ai_user_id_helpers.py
+++ b/tests/ai_user_id_helpers.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from contextlib import contextmanager
 from copy import deepcopy
+from dataclasses import replace
 from types import SimpleNamespace
 from typing import TYPE_CHECKING, TypeVar, cast
 from unittest.mock import AsyncMock, MagicMock
@@ -37,7 +38,7 @@ from mindroom.hooks.registry import HookRegistryState
 from mindroom.knowledge.utils import KnowledgeAvailabilityDetail, _KnowledgeResolution
 from mindroom.matrix.cache.thread_history_result import thread_history_result
 from mindroom.message_target import MessageTarget
-from mindroom.post_response_effects import PostResponseEffectsSupport
+from mindroom.post_response_effects import PostResponseEffectsDeps, PostResponseEffectsSupport
 from mindroom.response_payload_preparation import ResponsePayloadPreparer
 from mindroom.response_runner import (
     ResponseRequest,
@@ -56,7 +57,7 @@ from tests.conftest import (
 from tests.identity_helpers import persist_entity_accounts
 
 if TYPE_CHECKING:
-    from collections.abc import Generator, Iterable
+    from collections.abc import Callable, Generator, Iterable
     from pathlib import Path
 
     from agno.knowledge.knowledge import Knowledge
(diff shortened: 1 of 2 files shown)


basnijholt:
Re CodeRabbit's optional `dataclasses.fields()` copy in `_install_inert_post_response_effects`: skipping. A new required field on `PostResponseEffectsSupport` fails loudly here (TypeError on construction); a new defaulted field only matters if production starts reading it directly off the support, at which point tests using the inert swap fail visibly. The explicit five-field copy is clearer for a test helper than field-iteration metaprogramming, and the reviewer itself tagged the suggestion trivial/low-value.

PR #1425, opened 2026-07-08
Surface real provisioning error details instead of always claiming revoked credentials

basnijholt:
## Problem

`register_user_via_provisioning_service()` collapsed **every** 401/403 from `POST /v1/local-mindroom/register-agent` into:

> Provisioning credentials are invalid or revoked. Run `mindroom connect --pair-code ...` again.

But the server returns distinct 403s — notably the namespace enforcement added in #1038 (`{"detail": "Requested username is outside this local connection namespace"}`). Real-world debugging of that case wasted hours because the client claimed valid credentials were revoked.

## Changes

- Parse the JSON error body and extract `detail` when present, falling back to `response.text` truncated to 300 chars.
- Keep the "invalid or revoked / re-pair" message **only** for genuine client-auth failures: any 401, or a 403 whose detail matches the known auth strings from `scripts/local_mindroom_provisioning_service.py` (`Missing local client credentials`, `Invalid local client credentials`, `Connection revoked`).
- Other 403s raise a permanent startup error that includes the server's detail verbatim. The namespace-mismatch case additionally gets a hint that usernames must match `mindroom_<entity>_<namespace>` for this connection.
- The client secret and password never appear in any error text (asserted in tests).
- 404 and non-4xx branches keep their existing messages (the generic branch now benefits from JSON `detail` extraction instead of raw body text).
- The error branching moved into a `_raise_for_register_agent_error()` helper to keep the main function under the complexity limit.

## Tests

Extended `tests/test_matrix_agent_manager.py` with: 401 invalid client, 403 connection revoked, 403 namespace mismatch (asserts the detail and hint are surfaced and no secrets leak), and a 403 with a malformed (non-JSON) body that surfaces the raw text. The old fake-response 403 test was replaced by these, using real `httpx.Response` objects.

`uv run pytest tests/test_matrix_agent_manager.py -n 0 --no-cov`: 40 passed. `pre-commit` on both files: all hooks pass.

### DIFF
--- a/scripts/local_mindroom_provisioning_service.py
+++ b/scripts/local_mindroom_provisioning_service.py
@@ -53,6 +53,10 @@ RATE_LIMIT_STALE_SECONDS = 3600
 NAMESPACE_ALPHABET = "abcdefghjkmnpqrstuvwxyz23456789"
 NAMESPACE_LENGTH = 8
 MANAGED_AGENT_USERNAME_PREFIX = "mindroom_"
+# The local MindRoom client (src/mindroom/matrix/provisioning.py) classifies
+# errors by these exact strings; a contract test keeps the two sides in sync.
+CONNECTION_REVOKED_DETAIL = "Connection revoked"
+NAMESPACE_MISMATCH_DETAIL = "Requested username is outside this local connection namespace"
 PAIR_STATUS_SESSION_HEADER = "X-Local-MindRoom-Pair-Session-Id"
 
 
@@ -513,7 +517,7 @@ def _require_local_client(
     if not hmac.compare_digest(expected_hash, provided_hash):
         raise HTTPException(status_code=401, detail="Invalid local client credentials")
     if connection.revoked_at:
-        raise HTTPException(status_code=403, detail="Connection revoked")
+        raise HTTPException(status_code=403, detail=CONNECTION_REVOKED_DETAIL)
     return connection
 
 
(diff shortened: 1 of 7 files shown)


basnijholt:
Follow-up commit 1a60982 addresses self-review findings:

- **Password redaction**: error-detail extraction moved to a shared `mindroom.http_error_detail` module (also replacing `cli/connect.py`'s divergent copy). FastAPI 422 validation lists are summarized to `loc`/`msg` only — the request body echoed under `input` (which contains the agent password) can no longer reach error messages or startup logs. Reproduced the leak before fixing; a now-non-vacuous test posts a 422 body echoing the password and asserts it never appears in the raised error.
- **No more infinite retries on deterministic errors**: 400 (homeserver mismatch), 422 (schema skew), and 3xx redirects now raise permanent startup errors with the server detail, matching the sibling registration flows in `users.py`.
- **Contract test**: the two 403 detail strings are hoisted into named constants in `scripts/local_mindroom_provisioning_service.py`, and a test asserts the client sentinels match, so server rewording can't silently degrade classification. The dead 401-only entries in the client's detail set are gone (`_CONNECTION_REVOKED_DETAIL` is the only 403 auth detail the server sends).
- **Message polish**: 401 errors now include the server detail (diagnosable when a proxy, not the service, answers 401); the namespace hint has proper punctuation and points at the actual remedy (check `MINDROOM_NAMESPACE` / re-pair).

PR #1447, opened 2026-07-08
Make avatar generation robust to truncated prompts and no-image responses

basnijholt:
## Problem

`mindroom avatars generate` intermittently fails with `No image data found for teams/<name>` even when the API key and models are fine. Two compounding causes:

1. `_generate_prompt` caps the prompt model at `max_output_tokens=150`. Team prompts include per-member visual breakdowns and routinely exceed that, so the visual-elements text gets cut off mid-word (observed: a prompt ending in `... symmetry, brushed-slate texture\"\n*`). 
2. `_generate_avatar` makes exactly one image call. The image model occasionally responds to such mangled markdown-fragment prompts with text instead of an image (verified live: the same target failed twice, then produced an image on a later attempt with a freshly sampled prompt), and that single miss fails the entire generation run.

## Change

- Raise the prompt budget to 400 tokens (`_PROMPT_MAX_OUTPUT_TOKENS`).
- Retry the image call up to 3 attempts (`_MAX_IMAGE_ATTEMPTS`), regenerating the prompt on each attempt — the prompt is sampled at temperature 0.7, so retries are not identical — with a console note and a structured warning per miss. The final failure message now includes the attempt count.

## Testing

- New: retry succeeds on the second attempt with a fresh prompt (asserts both the prompt model and image model were called twice and the file content is the retried image).
- New: persistent no-image responses raise after exactly `_MAX_IMAGE_ATTEMPTS` calls and write nothing.
- Full `tests/test_avatar_generation.py` suite passes (31 tests); pre-commit clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved avatar creation reliability by retrying image generation when no image is returned.
  * Added clearer feedback when avatar generation fails after multiple attempts.
  * Limited prompt output length to keep generation more consistent.

### DIFF
--- a/src/mindroom/avatar_generation.py
+++ b/src/mindroom/avatar_generation.py
@@ -38,6 +38,14 @@ logger = get_logger(__name__)
 _PROMPT_MODEL = GOOGLE_AVATAR_PROMPT
 _IMAGE_MODEL = GOOGLE_AVATAR_IMAGE
 _ROOT_SPACE_AVATAR_NAME = "root_space"
+# Team prompts include per-member breakdowns; a low cap truncates them
+# mid-sentence and the mangled prompt makes the image model answer with
+# text instead of an image.
+_PROMPT_MAX_OUTPUT_TOKENS = 400
+# The image model occasionally returns no image for a valid prompt; each
+# retry regenerates the prompt (sampled at temperature), so attempts are
+# not identical.
+_MAX_IMAGE_ATTEMPTS = 3
 
 _ROOM_PURPOSES = {
     "lobby": "Central meeting space, entrance and welcome area",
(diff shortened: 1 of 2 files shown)


basnijholt:
Addressed the review items:

- The "regenerating prompt" console line now only prints when another attempt actually follows; the final miss goes straight to the failure message (greptile, qodo).
- No-image responses now log finish reasons plus a bounded text snippet (`_no_image_diagnostic`), so safety blocks are distinguishable from transient misses (qodo).
- On the ignore comments (gemini): `ty` reports both the mypy-style and the suggested `ty: ignore[arg-type]` forms as unknown/unused for these lines — it does not flag the fake-client arguments at all — so I removed the comments entirely rather than converting them.

Issue #1436, opened 2026-07-08
Trim the sandbox runtime import graph with lazy provider-SDK imports in the tool registry


basnijholt:
PR #1433 removed the per-call cost of the 11-17 s runtime import for worker-routed sandbox tools by paying it once in a warm forkserver template, but the import graph itself is unchanged and still taxes every template warm-up, pod boot, primary-runtime startup, and cooldown retry.

Measured on a staging worker pod (2-CPU cgroup limit, 2026-07-08):
- `import mindroom.api.sandbox_runner` alone: 11-17 s (bare interpreter: 0.03 s), CPU-bound pydantic model construction across LLM SDKs pulled in transitively by the tool registry.
- `-X importtime` worst offenders: `google.genai.types` 1.1 s; `mcp.types`, `fastapi.openapi.models`, `openai.*`, `anthropic.*`, `ollama.*`, `agno.*`, `nio.*` each 0.05-0.25 s, times hundreds of modules.
- Ceiling for this work: a perfectly slim file-tool entry still costs 3-4 s, because `mindroom.tools.file` imports 3.6 s of which `tool_system/metadata.py` alone is 3.2 s - so registry metadata needs to stop importing provider SDKs eagerly, not just the tool modules.

Goal: importing the registry (metadata + catalog) should not import any provider SDK; SDKs load on first tool construction. Benefits apply everywhere, not just sandbox dispatch: faster `mindroom run` startup, faster forkserver template warm-up, faster CI collection.

PR #1450, opened 2026-07-08
Stop the primary runtime from importing every provider SDK

basnijholt:
Extends #1442 (#1436) to the `mindroom run` startup path.

## Problem

#1442 fixed the slim entry points (registry, config, sandbox runner), but `import mindroom.orchestrator` — the bulk of `mindroom run` startup — still loaded **every** provider SDK (anthropic, google.genai, openai, ollama, groq, cerebras, deepseek) regardless of which providers the config uses. Four modules imported agno model classes at module top purely for `isinstance` dispatch, and the embedder factory imported both embedder SDKs eagerly:

| Module | Eager imports | Why they existed |
|---|---|---|
| `media_fallback.py` | 11 model classes | endpoint-attribute dispatch in `_route_endpoint` |
| `openai_tool_search.py` | `OpenAIResponses` | one `isinstance` gate |
| `thread_summary.py` | `VertexAIClaude` | one `isinstance` gate |
| `history/summary_call.py` | `Claude` | compaction model tuning gates |
| `embedding_factory.py` / `embeddings.py` | `OllamaEmbedder`, openai SDK | embedder construction + `MindRoomOpenAIEmbedder` subclassing agno's `OpenAIEmbedder` at class-definition time |

## Change

- New leaf helper `model_instance_checks.isinstance_of_loaded(instance, *(module, class))`: probes `sys.modules` for the class's concrete defining module and runs `isinstance` only against loaded classes. This is semantically identical to importing the class — an instance (including subclasses like `MindroomVertexAIClaude` or `CodexResponses`) can only exist once its module is imported — the same invariant #1442 established for `claude_prompt_cache.as_anthropic_claude`. Plain string dispatch on class names would break exactly those subclass cases, which is why `isinstance` stays.
- The four dispatch modules switch to it, with `TYPE_CHECKING` imports + casts keeping the narrowed attribute access typed.
- `MindRoomOpenAIEmbedder` moves to its own module (`openai_embedder.py`); `embeddings.py` keeps the light signature helpers its other importers (memory config, knowledge indexing) actually need; the embedder factory imports each provider's embedder inside its branch.
- The import-graph regression test grows a `mindroom.orchestrator` probe (provider SDKs only — `nio` and `mcp` are genuine runtime-boot dependencies: the MCP manager syncs unconditionally at startup).

## Measurements (warm `-X importtime`, M1-class Mac, best of 3)

| | main | PR |
|---|---|---|
| `import mindroom.orchestrator` | 2.60 s, loads 7 provider SDKs | **1.90 s, loads none** |

The configured provider's SDK still loads at first model construction during startup (correct), so a single-provider config now pays for one SDK instead of seven. Remaining known weight, deliberately left: `mem0` eagerly imports `qdrant_client` (~0.32 s) upstream even though we use chroma, and mem0 loads at startup whenever memory is enabled; `nio`/`mcp`/`chromadb`/agno core are genuine runtime dependencies.

## Testing

- Full suite: 9361 passed, 61 skipped.
- `uv run pre-commit run --all-files` clean; `tach check --dependencies --interfaces` clean.
- New probe fails on main, passes here.

## Summary by CodeRabbit

* **New Features**
  * Added support for OpenAI-compatible embedding providers with configurable dimensions and robust single/batch embedding + usage handling.

* **Bug Fixes**
  * Improved provider/model-type detection for summary configuration, media routing, and tool-search so the right provider-specific behavior is applied.
  * Reduced unnecessary provider SDK loading via deferred and import-safe runtime checks.

* **Tests**
  * Strengthened import-safety regression coverage and added tests for “loaded-only” class detection behavior.

### DIFF
--- a/src/mindroom/claude_prompt_cache.py
+++ b/src/mindroom/claude_prompt_cache.py
@@ -49,10 +49,10 @@ model with no deferred tools can still replay poisoned history.
 
 from __future__ import annotations
 
-import sys
 from typing import TYPE_CHECKING, Any, cast
 
 from mindroom.model_defaults import TOOL_SEARCH_UNSUPPORTED_MODEL_ID_PREFIXES
+from mindroom.model_instance_checks import isinstance_of_loaded
 
 if TYPE_CHECKING:
     from agno.models.anthropic import Claude as AnthropicClaude
@@ -89,15 +89,17 @@ def native_tool_search_supported(provider: str, model_id: str) -> bool:
     return not model_id.startswith(TOOL_SEARCH_UNSUPPORTED_MODEL_ID_PREFIXES)
 
 
+_ANTHROPIC_CLAUDE_CLASS = ("agno.models.anthropic.claude", "Claude")
+
+
 def as_anthropic_claude(model: object) -> AnthropicClaude | None:
     """Narrow one model to Agno's Claude without importing the anthropic SDK.
 
-    A Claude instance can only exist once ``agno.models.anthropic`` is
-    imported, so probing ``sys.modules`` keeps non-Claude runtimes from paying
+    A Claude instance can only exist once ``agno.models.anthropic.claude`` is
+    imported, so the loaded-class check keeps non-Claude runtimes from paying
     the anthropic import just to no-op these hooks (#1436).
     """
-    module = sys.modules.get("agno.models.anthropic")
-    if module is None or not isinstance(model, module.Claude):
+    if not isinstance_of_loaded(model, _ANTHROPIC_CLAUDE_CLASS):
         return None
     return cast("AnthropicClaude", model)
 
--- a/src/mindroom/embedding_factory.py
+++ b/src/mindroom/embedding_factory.py
@@ -4,10 +4,8 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from agno.knowledge.embedder.ollama import OllamaEmbedder
-
 from mindroom.credentials_sync import get_api_key_for_provider, get_ollama_host
-from mindroom.embeddings import MindRoomOpenAIEmbedder, create_sentence_transformers_embedder
+from mindroom.embeddings import create_sentence_transformers_embedder
 from mindroom.model_defaults import OLLAMA_HOST_DEFAULT
 
 if TYPE_CHECKING:
@@ -23,6 +21,10 @@ def create_configured_embedder(config: Config, runtime_paths: RuntimePaths) -> E
     embedder_config = config.memory.embedder.config
 
     if provider == "openai":
+        # Imported at first construction so only the configured embedder
+        # provider's SDK loads (#1436).
+        from mindroom.openai_embedder import MindRoomOpenAIEmbedder  # noqa: PLC0415
+
         return MindRoomOpenAIEmbedder(
             id=embedder_config.model,
             api_key=get_api_key_for_provider("openai", runtime_paths=runtime_paths),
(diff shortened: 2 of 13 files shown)


basnijholt:
All four review findings addressed in a3d36e4c7 — each verified against the tree before fixing:

1. **Tach boundaries** — confirmed: the clean check was an artifact of the two new modules being undeclared. `mindroom.model_instance_checks` (`depends_on = []`) and `mindroom.openai_embedder` (`depends_on = ["mindroom.model_defaults"]`) are now declared alongside the other leaf modules, and `embedding_factory`, `history.summary_call`, and `thread_summary` list their real new edges. `tach check --dependencies --interfaces` passes with the imports visible.
2. **Probe consolidation** — `as_anthropic_claude` is now a thin narrowing wrapper over `isinstance_of_loaded` using the concrete defining module (`agno.models.anthropic.claude`), and `summary_call` reuses it, dropping its own tuple constant and both `cast("Claude", ...)` sites. One probe implementation, one path convention.
3. **Tuple guard tests** — new `tests/test_model_instance_checks.py`: a parametrized test walks all 12 declared `(module, class)` tuples (media_fallback ×10, openai_tool_search, thread_summary, claude_prompt_cache) and asserts `getattr(import_module(mod), cls).__module__ == mod` — the defining-module assertion also catches package-init re-exports, not just typos. Plus unit tests for subclass matching, unloaded-module no-match, and multi-path tuples. Thread_summary's inline tuple was hoisted to a module constant so the guard can see it.
4. **Dead aliases** — dropped both `_OPENAI_EMBEDDING_DIMENSIONS` and the equally pointless `_DEFAULT_SENTENCE_TRANSFORMERS_MODEL` from `embeddings.py`; call sites use the `model_defaults` constants directly.

Full suite: 9375 passed, 61 skipped; `pre-commit run --all-files` clean.

basnijholt:
7e4427d90 replaces the class-string dispatch in `media_fallback` with structural dispatch: `_route_endpoint` now asks which endpoint attribute the model carries (`azure_endpoint` / `host` / `base_url` / `client_params`) via `runtime_checkable` Protocols and plain `isinstance` — the same idiom `thread_summary._SupportsTemperature` already used. Ten of the twelve `(module, class)` string tuples, all the casts, and the Vertex-before-Claude ordering comment are gone (the ordering falls out of the structure). `thread_summary`'s Vertex special case is now a plain comparison against agno's own `provider` label.

Equivalence verified empirically: old and new dispatch agree on all fifteen relevant model classes (including `LlamaCpp`, `OpenAILike`, `CodexResponses`, `MindroomVertexAIClaude`) with and without endpoints set.

The `sys.modules` probe survives only in the two identity-gated wire hooks (`as_anthropic_claude`, the `OpenAIResponses` check in `openai_tool_search`), where structural dispatch can't express "this exact client family" and a call-time import would reload every SDK; the tuple guard test now covers exactly those two.

Full suite 9366 passed / 61 skipped; pre-commit and tach clean; net −23 lines.

basnijholt:
Good catch on the `thread_summary` provider-label check — addressed in the latest commit.

Verified the current blast radius first: today only the Vertex Claude family carries `provider == "VertexAI"` (agno's `Gemini` keeps `provider="Google"` even with `vertexai=True`, and `agno.models.vertexai` contains only `claude`), so the over-match wasn't reachable yet. But the rawPredict temperature restriction is specific to Vertex *Claude*, and the label is a looser predicate than the class check it replaced — an agno release adding any other VertexAI-labeled model would have silently nulled its configured temperature.

Temperature suppression is an identity question (like the anthropic and OpenAI Responses wire hooks), not a shape question (like the media-route dispatch), so it's back on `isinstance_of_loaded(model, _VERTEXAI_CLAUDE_CLASS)` — exact class family, subclasses included — and the tuple guard test covers it again. Full suite 9367 passed; pre-commit and tach clean.

PR #1500, opened 2026-07-13
Add bulk thread-cache backfill: one room scan warms many dormant threads

basnijholt:
## Problem, with production evidence

Backfilling thread exports for a large dormant room on the mindroom-chat instance crawled to 1-2 threads/minute. Measured during the pass:

```
homeserver_fetch_ms: 85337.6, homeserver_scan_pages: 119, homeserver_scanned_event_count: 11900
homeserver_fetch_ms: 70005.6, homeserver_scan_pages: 119, homeserver_scanned_event_count: 11900
homeserver_fetch_ms: 55177.7, homeserver_scan_pages: 120, homeserver_scanned_event_count: 12000
```

Each dormant thread's refresh walks room history backward from the newest event until it sees that one thread's root, then throws the walk away. For threads rooted ~12k events deep that is 120 paginated calls and 50-85 seconds per thread, and the next thread repeats the same walk. For a room with H events and T dormant threads the total is O(T x H) homeserver work, plus rate-limit backoff (41 hits in one 30-minute window). The runtime never notices because dispatch only fetches one thread at a time; a bulk backfill is the pathological case.

## Fix

New `bulk_refresh_room_thread_histories(client, room_id, event_cache, *, thread_root_ids)` in `client_thread_history`:

- One backward walk of the room (same `room_messages` filter, stops early once every requested root has been seen).
- One canonical resolution pass (`resolve_thread_ids_for_event_infos`) buckets every scanned event by thread; per-thread rows are root-first sorted and carry the same latest-edit sidecars as the per-thread scan.
- Each requested thread is stored through the same guarded `replace_thread_if_not_newer` path with the scan's start time, so the store-race guard (PR #716) and root-required cacheability rule (PR #741) apply unchanged.
- Roots the drained scan never saw are returned in `missing_root_ids` and never stored (no partial snapshots).

The thread exporter uses it automatically in `prefer_cache` mode: after enumerating a room it probes which threads the durable cache would refuse to serve (`untrusted_cached_thread_ids`, same `thread_cache_rejection_reason` gate as reads), and when 10 or more would miss, it warms them with one scan. Scan-proven-missing roots are recorded as failures without paying one full per-thread walk each. Any bulk failure logs and falls back to the per-thread path unchanged.

## Why it fixes it

The same production room: ~700 threads x up to 130 pages each becomes one ~130-page walk plus SQLite reads. Homeserver work drops from O(T x H) to O(H), which also removes almost all of the rate-limit pressure. Reconstruction semantics are identical by construction: the bulk path reuses the existing scan recorder, resolution, ordering, and store primitives rather than reimplementing them.

## Testing

- `tests/test_bulk_thread_backfill.py`: one scan stores multiple threads root-first with correct bucketing; early stop after all roots found; drained scans report missing roots and store no partial threads.
- `tests/test_thread_export_execution.py`: prefer_cache passes with >= 10 untrusted threads trigger exactly one bulk refresh and fail missing roots without per-thread fetches; below the threshold the bulk path is skipped entirely.
- Existing per-thread scan behavior is untouched (`_record_scanned_room_message_source` refactor is signature-only); full thread-domain and export suites pass, pre-commit clean, new API exposed through the tach interface.

## Summary by CodeRabbit

* **Improvements**
  * Thread exports can now refresh multiple cached conversation threads using a single backward room scan, reducing work for large exports.
  * When cache trust is enabled, untrusted cached thread roots are proactively warmed in bulk; otherwise the exporter falls back to per-thread retrieval.
  * Missing thread roots are reported and excluded so partial/unreliable histories are not exported.
* **New Features**
  * Added bulk refresh statistics reporting for requested, stored, missing roots, scan pages, and scanned event count.
* **Tests**
  * Added and updated tests covering bulk cache refresh and cache-preference behavior during thread export execution.

### DIFF
--- a/src/mindroom/matrix/client_thread_history.py
+++ b/src/mindroom/matrix/client_thread_history.py
@@ -33,6 +33,7 @@ Cache-trust rules (each encodes a shipped regression fix; do not weaken them):
 
 from __future__ import annotations
 
+import asyncio
 import time
 from collections.abc import Collection, Iterable, Mapping, Sequence
 from dataclasses import dataclass
@@ -1164,6 +1165,13 @@ async def _group_scanned_sources_by_thread(
     latest_edits_by_original_event_id: dict[str, tuple[nio.RoomMessageText | nio.RoomMessageNotice, str | None]],
 ) -> dict[str, list[dict[str, Any]]]:
     """Bucket one room scan's event sources per requested thread with one canonical resolution."""
+    grouped: dict[str, dict[str, dict[str, Any]]] = {
+        root_id: {root_id: scanned_message_sources[root_id]}
+        for root_id in thread_root_ids
+        if root_id in scanned_message_sources
+    }
+    if not grouped:
+        return {}
     event_infos = {
         event_id: EventInfo.from_event(event_source) for event_id, event_source in scanned_message_sources.items()
     }
@@ -1173,11 +1181,6 @@ async def _group_scanned_sources_by_thread(
         event_infos=event_infos,
         ordered_event_ids=ordered_event_ids,
     )
-    grouped: dict[str, dict[str, dict[str, Any]]] = {
-        root_id: {root_id: scanned_message_sources[root_id]}
-        for root_id in thread_root_ids
-        if root_id in scanned_message_sources
-    }
     for event_id in ordered_event_ids:
         root_id = resolved_thread_ids.get(event_id)
         if root_id is None or root_id == event_id:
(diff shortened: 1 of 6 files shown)


basnijholt:
Addressed the review findings in ffb997fd5:

- **Bulk edits not re-sorted (qodo, action required): fixed.** Verified real: the per-thread scan's final step re-sorts after appending edit sources; the bulk path sorted before appending. Buckets are now sorted root-first after the edit sources are attached, matching the per-thread ordering exactly, and the bulk test now includes an edit event asserting the final order.
- **O(threads x edits) edit attachment (gemini): fixed.** Edits are pre-grouped by resolved thread root in one pass over the edit map.
- **Resolution runs when no roots found (qodo, optional): fixed.** The room-wide resolution pass is skipped when the scan contains none of the requested roots.
- **Sequential cache-state probes (gemini high, greptile): fixed.** untrusted_cached_thread_ids now gathers all get_thread_cache_state calls concurrently.
- **Threshold is scan-depth-blind (greptile): fixed, by going further.** The observation is right that a count threshold ignores depth, but the bulk scan stops as soon as every requested root has been seen, so its cost is one walk to the deepest requested root while the per-thread path pays one walk per thread. That makes bulk favorable from two threads up regardless of depth; threshold lowered from 10 to 2 with the reasoning documented at the constant.
- **Duplicate cache-state reads between probe and per-thread fetch (qodo, recommended): declined.** The second read is the trusted read path's own gate; removing it would mean plumbing pre-read state into fetch_thread_history and weakening the seam that owns cache trust, to save cheap point lookups. Not worth the coupling.

basnijholt:
Self-review follow-up in d4b17cd84 (net -91 lines): the per-thread scan is now a thin wrapper over the bulk walk with a single requested root, deleting the duplicated page-walk loop and the one-thread filter helper (the canonical-membership regression tests moved to the shared grouping seam). Also dropped the untrusted-count threshold entirely: since the scan early-stops at the deepest requested root, bulk costs one deep walk regardless of thread count, so the exporter now bulk-warms whenever any enumerated thread would miss the cache.

PR #1544, opened 2026-07-15
Keep Vertex tool search out of token-count requests

basnijholt:
## Why

Vertex Claude generation supports native Tool Search, but Vertex’s token-count endpoint rejects the Tool Search server type and `defer_loading` metadata. The new context guard sent the generation payload to token counting unchanged, causing valid Fable requests to fail before generation.

## What changed

Token counting now removes only the unsupported server-tool entry and `defer_loading` annotations. It retains every custom tool definition, producing a conservative count while generation keeps native deferred loading and its context-efficiency benefits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token counting for Vertex AI requests that use tool-search features.
  * Ensured generation-only tool metadata is excluded from token calculations while preserving relevant tool schema information.
  * Improved accuracy of reported input token totals.

### DIFF
--- a/docs/tools/dynamic-tools.md
+++ b/docs/tools/dynamic-tools.md
@@ -32,8 +32,7 @@ No flags means the tool is eager and appears in every request.
 
 ## Native Server-Side Tool Search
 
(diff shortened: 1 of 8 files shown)


basnijholt:
Closing in favor of #1545, which handles Vertex tool-search replay history and avoids counting the full deferred-tool catalog. #1545 still needs to count schemas for tools discovered in prior search results before merge; its fixed reserve alone can undercount those expanded definitions.

PR #1597, opened 2026-07-20
Preserve requester identity during restart auto-resume

basnijholt:
## Why

A restart-interrupted response is resumed through a router-authored relay. The relay preserved the human in `com.mindroom.original_sender`, but omitted the trusted relay source kind, so ingress rejected that requester metadata and treated the router bot as the requester. Private execution then selected a new bot-keyed workspace with the wrong memory and OAuth bindings.

Requester recovery is also best-effort after restart. If the original requester cannot be recovered, continuing as the bot is unsafe.

## What changed

- Mark auto-resume messages as `trusted_internal_relay` whenever they carry `com.mindroom.original_sender`.
- Skip auto-resume when restart recovery cannot resolve the requester identity.
- Cover relay content, ingress requester promotion, and the existing managed-bot trust boundary.

## Verification

- `uv run pytest tests/test_stale_stream_cleanup.py tests/test_ingress_validation.py -q`
- Full suite reached 10,844 passing tests. PostgreSQL integration tests could not start because local Docker storage was full; unrelated timing-sensitive failures passed on immediate isolated rerun.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic conversation resumption by preserving the original sender’s identity when messages are relayed internally.
  - Prevented auto-resume messages from being sent when the original requester cannot be verified.
  - Strengthened validation so embedded sender identities are trusted only for verified internal relay messages.
  - Added safeguards to prevent untrusted external messages from impersonating another sender.
- **Tests**
  - Expanded coverage for sender verification and stale-stream auto-resume scenarios.

### DIFF
--- a/src/mindroom/dispatch_source.py
+++ b/src/mindroom/dispatch_source.py
@@ -25,6 +25,9 @@ HOOK_DISPATCH_SOURCE_KIND = "hook_dispatch"
 EXTERNAL_TRIGGER_SOURCE_KIND = "external_trigger"
 ACTIVE_THREAD_FOLLOW_UP_SOURCE_KIND = "active_thread_follow_up"
 TRUSTED_INTERNAL_RELAY_SOURCE_KIND = "trusted_internal_relay"
+AUTO_RESUME_MESSAGE = (
+    "[System: Previous response was interrupted by service restart. Please continue where you left off.]"
+)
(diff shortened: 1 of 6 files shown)


basnijholt:
@coderabbitai review

PR #1615, opened 2026-07-21
feat: publish cron descriptions in scheduled task state

basnijholt:
## Problem

MindRoom already owns cron parsing and human-readable schedule descriptions, but scheduled-task Matrix state only exposed the serialized workflow.
MindRoom Chat therefore had to reconstruct scheduler semantics in the browser to show recurring timing.

## Change

- Publish the existing backend-generated `cron_description` as a top-level scheduled-task state field.
- Keep the field null for one-time schedules.
- Cover the Matrix persistence contract with a focused test.

## Validation

- `uv run pytest tests/test_scheduling.py -x -n 0 --no-cov -q`
- Full pytest reached 11,324 passing tests; six unrelated Git metadata and shell-buffer timing tests failed under the parallel full-suite load and all six passed together in an isolated serial rerun.

## Related

Paired with mindroom-ai/mindroom-chat#188, which consumes this field and removes client-side cron evaluation.

### DIFF
--- a/src/mindroom/scheduling.py
+++ b/src/mindroom/scheduling.py
@@ -746,6 +746,9 @@ async def _persist_scheduled_task_state(
         content={
             "task_id": task_id,
             "workflow": workflow.model_dump_json(),
+            "cron_description": (
+                workflow.cron_schedule.to_natural_language() if workflow.cron_schedule is not None else None
+            ),
             "status": status,
             "created_at": _serialize_scheduled_task_created_at(created_at),
             "updated_at": datetime.now(UTC).isoformat(),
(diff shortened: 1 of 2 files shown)


basnijholt:
Review follow-up:

- Fixed in 81475a9af: `cron_description` is now gated by `schedule_type == "cron"`, with regression coverage for a contradictory one-time workflow that also carries a cron field.
- The cancellation claim does not apply to this contract: cancelled state is deliberately non-pending and excluded by scheduled-task UI/listing consumers. `_cancelled_task_content` preserves the canonical workflow, from which backend read models derive descriptions when needed. Copying a pending-only display field into terminal state would add lifecycle coupling without changing behavior.

Validation: all 56 scheduling tests pass; pre-commit passes for both changed files.

PR #1627, opened 2026-07-22
Bound STT transcription with one total deadline

basnijholt:
STT requests currently inherit HTTPX's 5-second read timeout, which can reject healthy but slow transcription. This change gives transcription one attempt with a 60-second total deadline covering connection setup, upload, and response reading. Connection and pool waits remain limited to 5 seconds, and upload inactivity to 10 seconds.

Timeouts and service errors return to the existing audio-attachment fallback without resubmitting the audio. External cancellation propagates. The STT budget excludes Matrix media download and transcript normalization.

Validation:
- 73 focused tests passed across STT request handling, voice fallback, and ingress ordering, including cancellation and deadlines while waiting for response headers or body data.
- Pre-commit checks for all changed files passed, including typing, Tach boundaries, module privacy, and regenerated voice documentation.
- Independent code review found no blockers.

### DIFF
--- a/src/mindroom/voice_handler.py
+++ b/src/mindroom/voice_handler.py
@@ -44,6 +44,13 @@ if TYPE_CHECKING:
     from mindroom.constants import RuntimePaths
 
 logger = get_logger(__name__)
+# STT latency scales with clip length, so httpx's 5s default read timeout is never enough for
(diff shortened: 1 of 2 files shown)


basnijholt:
@coderabbitai full review

PR #1647, opened 2026-07-24
Reduce contention under concurrent MindRoom sessions

basnijholt:
## Why

Live profiling under only ten concurrent threads showed the application was not CPU-throttled, but synchronous maintenance and shared coordination were blocking the event loop and growing queues. Knowledge refreshes also remained unusable because legacy metadata looked config-incompatible, while file indexing ran with fixed low concurrency.

## What changed

- use raw Kubernetes Deployment snapshots for every list caller, and share one list between cleanup and reconciliation
- expose typed worker maintenance through the backend protocol and `WorkerMaintenanceResult`
- move synchronous session-history cleanup off the event loop with worker-owned storage handles
- batch handled-turn persistence with one ordered drain per ledger, bounded cross-ledger parallelism, exact durability waiters, and finite failure propagation
- share Matrix typing leases across concurrent turns, including safe stop/reacquire and best-effort failure handling
- scope event-cache coordination per principal instead of serializing unrelated agents
- bound optional streaming delivery requests once the pending queue reaches 32 while preserving required capture and refresh work
- normalize legacy optional knowledge filters at the metadata parse boundary
- validate and document configurable knowledge file indexing concurrency from 1 to 128

## Evidence

- the old typed Kubernetes list measured 0.896 s wall / 0.874 s CPU per call; the raw snapshot path measured 0.057 s / 0.042 s
- all worker-list callers now use the raw path; maintenance also collapses two list calls into one, reducing that polling CPU path by roughly 42x
- 10-way and 100-way concurrency tests cover typing lease sharing, event-cache scheduling, ledger batching and retry exhaustion, and stream bounding
- the full Python CI suite, smoke deployment, container builds, plugin checks, scans, and repository hooks pass

## Review follow-ups

Fixed in `8149c93a1` after review:

- **Typing lease start/refresh failures no longer stick.** The shared refresh loop previously ended on the first failed Matrix request, so a turn joining that lease afterwards never saw typing and no retry ran. Because the lease outlives individual turns, this converted a transient blip into a sticky per-room outage; before the lease consolidation each turn retried independently. Now covered by a joiner-recovery test.
- **Kubernetes `status` is optional.** It is written by the deployment controller, so a Deployment read before that first write has none; requiring it aborted cleanup and reconciliation for every worker in the pass. `metadata.labels` is likewise no longer required — the snapshot carries labels but never reads them.
- **`WorkerMaintenanceResult` uses tuples.** `frozen=True` did not prevent mutation through the previous list fields.
- **Removed defensive defaults** on the private reconcile helpers (`scaled_down`, `scaled_down_worker_ids`).
- **Documented the knowledge filter producer invariant** behind the legacy optional-key normalization, including why `skip_hidden` is deliberately left un-normalized.

Note on scope: this PR bundles several independent subsystem changes (Kubernetes listing, worker maintenance protocol, session-history threading, ledger batching, typing leases, cache scoping, stream bounding, knowledge filters, indexing concurrency). The typing-lease change was the highest-risk piece and would have benefited from separate review.


basnijholt:
@coderabbitai review

basnijholt:
Addressed both Qodo findings in 211b3bae3: queued-notice cleanup now creates and closes a fresh storage handle inside the worker thread, and coordinator lanes now use structured tuple keys. Added worker-thread ownership and key-collision regressions; focused suites and all pre-commit hooks pass.

PR #1649, opened 2026-07-25
Recover Matrix cache after cancelled Postgres work

basnijholt:
## Why

A cancelled Postgres cache operation can leave its psycopg async connection busy. The transient classifier did not recognize that state, so later cache work kept reusing the poisoned connection and fail-closed invalidation could disable the advisory cache. Startup prewarm then continued fetching snapshots even though none could persist.

## What

- Treat the psycopg connection-busy error as transient so cache work reconnects and retries.
- Skip or abort startup thread prewarm when durable cache writes are unavailable.
- Report writes_unavailable separately from a guarded replacement race.
- Cover reconnect/retry and prewarm-skip behavior with regression tests.

## Validation

- uv run --extra postgres pytest -x --lf tests/test_event_cache_backends.py tests/test_bot_ready_hook.py
- 43 passed, 2 skipped
- Commit-time Python quality, type, boundary, and privacy hooks passed

### DIFF
--- a/src/mindroom/matrix/cache/postgres_event_cache.py
+++ b/src/mindroom/matrix/cache/postgres_event_cache.py
@@ -57,6 +57,7 @@ _TRANSIENT_SQLSTATES: frozenset[str] = frozenset(
     },
 )
 _TRANSIENT_ERROR_TEXT: tuple[str, ...] = (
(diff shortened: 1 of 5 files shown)


basnijholt:
@coderabbitai review

PR #1658, opened 2026-07-25
Use model context for compaction summaries

basnijholt:
## Summary

- keep `replay_window_tokens` scoped to persisted replay and compaction-trigger planning
- size summary input from the selected compaction model context, preserving reserve, prompt overhead, safety margin, and shrink retry
- update configuration docs and regression coverage

A 200k replay cap no longer forces a 1M-context summary model to split otherwise fitting history across several summary requests. With the default reserve, a 1M model gets an 881,616-token summary-input budget while replay remains capped at 200k.

## Tests

- `.venv/bin/pytest tests/test_history_replay_planning.py tests/test_compact_context.py tests/test_compaction_invariants.py -x --lf -n auto --no-cov -q`
- `.venv/bin/pytest -x --lf -n auto --no-cov -q`
- `.venv/bin/pre-commit run --files <changed files>`

### DIFF
--- a/docs/configuration/agents.md
+++ b/docs/configuration/agents.md
@@ -193,7 +193,7 @@ Crossing that soft trigger while still within the hard budget leaves the stored 
 
 You can tune compaction behavior with these settings:
 
-- Use `replay_window_tokens` to cap persisted replay, required-compaction planning, and summary input chunks below the model's real context window without lowering the provider request limit.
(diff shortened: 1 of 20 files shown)


basnijholt:
Addressed the Greptile P1 in 75c506957. The primary keeps its full model-context budget. A smaller fallback now gets its own resolved input budget, rebuilds the refused chunk only when necessary, and keeps that budget for later chunks. Added planning and multi-chunk regression coverage; full pytest and changed-file pre-commit pass.

PR #1662, opened 2026-07-25
Make large knowledge-index refresh resumable and bounded

basnijholt:
## Problem

A semantic knowledge refresh builds its vectors into a private *candidate* Chroma collection and publishes only when the candidate provably matches the live source. That part is correct, but the candidate itself was entirely disposable:

- `KnowledgeManager.reindex_all` created a **fresh** candidate collection on every refresh (`_candidate_collection_name` embedded a timestamp and a random suffix).
- All progress lived in process memory (`candidate_indexed_files`, `candidate_indexed_signatures`).
- The `finally` block deleted the candidate collection whenever `_CandidatePublishState.index_published` was false — which is *every* non-publishing outcome: one failed file, a source change detected during the trailing signature comparison, a cancellation, or a crash before the metadata write.

So the durable progress of an interrupted refresh was exactly zero, and the next refresh started a new candidate from scratch. Combined with one embedding request per chunk, one `asyncio.gather` over the entire file list, and one INFO log line per file, a corpus large enough to take longer to index than the interval between interruptions could never publish, while abandoned candidate collections accumulated in the vector store.

This restores the resumability that PR #738 removed, without giving up any of the atomic-snapshot invariants that PR added.

### Durable candidate state — `src/mindroom/knowledge/candidate_checkpoint.py`

A new schema-versioned checkpoint records everything needed to safely continue a build: candidate collection name, full `IndexingSettings` fingerprint (base/storage/source identity, mode, embedder provider/model/host/dimensions, chunking, repository and filter identity), target source revision, per-file content signatures, failed files with retry state, totals, and timestamps.

It is stored as a compacted snapshot (`candidate_index.json`, written atomically) plus an append-only journal (`candidate_index.jsonl`). Recording one completed file is one small append rather than a rewrite of a snapshot that grows with the corpus; loading replays the journal over the snapshot and ignores a torn trailing line from a crash. Compaction writes the snapshot first and then removes the journal, so replaying an already-compacted journal is idempotent.

Journaling happens **per file**, not per batch, so an interruption partway through a batch still keeps every file that finished.

### Resume — `src/mindroom/knowledge/manager.py`

`reindex_all` now opens a candidate run instead of creating one:

1. Load the persisted candidate. Discard it if its settings fingerprint does not match, if its collection is missing, or if it points at the currently published collection (a crash between publication and checkpoint cleanup).
2. Reclaim candidate collections abandoned by earlier crashed refreshes.
3. Reconcile against the current listing: drop entries for removed and changed paths (vectors and checkpoint entry together, so a later failure cannot strand stale vectors), verify remaining entries actually have vectors, and build the pending set from what is missing, changed, or previously failed.
4. Index only the pending files, journaling each result.
5. Recheck completeness against a freshly computed live source signature. On a mismatch, reconcile again rather than discarding — bounded to a few rounds so a source that changes on every pass keeps its progress for the next refresh instead of thrashing.
6. Publish, then retire the checkpoint and delete superseded collections.

The checkpoint is written *before* the candidate collection is created, so a crash can never strand a collection that nothing references.

### Source advancement

The candidate is content-signature based rather than frozen to one revision. Each refresh syncs the source, reconciles the candidate against the new listing, and re-embeds only the delta: unchanged files keep their vectors, changed files are re-embedded, deleted files have their vectors removed. `target_revision` advances only when the reconciled state it describes is being made durable. Refreshes for one source root already serialize behind a per-source lock and coalesce in the scheduler, so polling and manual refresh converge on one candidate instead of racing rebuilds.

### Retry — `src/mindroom/knowledge/index_retry.py`, `src/mindroom/embedding_errors.py`

`embedder_failure_is_transient` classifies transport faults, timeouts and HTTP 408/429/5xx as retryable, and authentication/permission rejections, malformed payloads and contract violations as permanent. `EmbedderRequestError` now carries the provider's `Retry-After` hint across the credential-redacting boundary, so backoff can honor it (clamped, with jitter). Permanent failures stop the pass immediately instead of issuing one doomed request per remaining file, and are reported through the same "Indexed N of M …" path as per-file failures.

Exhausting the retry budget preserves the candidate, records the failed file and its attempt count, marks the refresh failed, and leaves any last-good index queryable. The next refresh retries only the unresolved work.

### Batching — `src/mindroom/knowledge/embedding_batch.py`

Agno embeds one chunk per provider request (`ChromaDb._upsert` → `Document.embed` → `Embedder.get_embedding_and_usage`). Rather than reimplementing Agno's write path and losing its id/metadata/Chroma-batching semantics, a narrow adapter sits at the MindRoom boundary: `BatchPrefetchEmbedder` wraps the configured embedder and serves already-embedded chunk texts from a short-lived cache. The indexer reads and chunks a bounded batch of files, embeds those chunk texts in as few provider requests as the item and payload limits allow (64 items / 512 KB by default), then hands the files to Agno, whose per-chunk embed calls become cache hits.

Cache misses fall through unchanged, so behavior is identical for providers without a batch surface, for content that changed between planning and insertion, and for query-time embedding. A batch that exhausts its transient retries degrades to per-file embedding, which attributes the failure to specific files and never re-embeds anything already cached.

### Bounded work

Files are processed in bounded batches (64) with the existing per-file concurrency limit applied inside each batch, so live asyncio tasks no longer scale with corpus size. Chunked signature scanning keeps the event loop responsive, and vector verification is batched into `$in` queries.

### Observability

Per-file logging moved to DEBUG. INFO now carries periodic `knowledge_candidate_progress` summaries and one `knowledge_candidate_finished` summary with total / completed / pending / failed / retrying / target revision / whether the pass resumed / elapsed / throughput / approximate ETA. `KnowledgeIndexStatus` gained an optional `candidate` field, and the API responses an optional `candidate` object — additive, omitted entirely when no candidate exists, and deliberately kept separate from `indexed_count` so candidate work is never mistaken for a queryable index.

### Garbage collection

`_cleanup_superseded_collections` takes an explicit preserved set and a `candidates_only` mode used at refresh start. Ownership is proven by name (both the default collection name and the candidate prefix embed the base identity and a digest of the resolved source path, inside the base's own private storage directory). Anything else is preserved and reported rather than deleted. The published collection and the active candidate are never deleted, and at most one candidate per base is retained.

## Safety invariants preserved

- A partial candidate is never exposed as published.
- The last-good published index stays queryable for the whole refresh.
- Metadata publication and collection adoption remain atomic; cancellation at any boundary cannot produce a false-complete state, and a cancellation after the metadata write is recognised on the next run so the live index is never reopened for writing.
- A candidate is only resumed when settings, embedder identity, source identity and repository identity all match.
- Concurrent refreshes serialize on the existing per-source lock and share one candidate.
- Deletes and content changes remove their vectors from the candidate.
- Cross-base and cross-workspace reuse remains impossible: the settings fingerprint includes base ID, storage root and resolved source path.

## Tests

`tests/test_knowledge_resumable_refresh.py` (new) drives the lifecycle through a fake vector store and a recording embedder, so assertions about *which* texts were embedded and how many provider requests were issued are exact:

- Cold-start interruption resumes the same candidate and does not re-embed durably completed files.
- A failing candidate resumes privately while the published index stays queryable.
- A transient fault retries only the affected work; exhausting the budget keeps the candidate and resumes only unresolved files.
- Permanent failures never publish, keep last-good, report only classified text, and cost at most a couple of requests.
- Incompatible settings discard the candidate without touching the published index.
- Add/modify/delete during a build reuses unchanged work and removes deleted vectors; a change racing final verification reconciles instead of destroying.
- Concurrent refreshes share one candidate.
- Cancellation around publication, and a checkpoint pointing at the published collection.
- GC preserves the published collection, the active candidate and unknown collections; repeated interrupted refreshes keep the collection count at one.
- Batch planning honors item and payload limits; request count scales with batches; batch failure falls back to per-file without re-embedding successes; non-batching providers degrade safely.
- A 400-file corpus keeps live tasks bounded; a 128-file build emits no per-file INFO lines.
- Candidate progress cannot be confused with published counts, and disappears once published.
- A completed checkpoint entry without vectors is requeued.
- Legacy published metadata without any candidate fields stays queryable and is not rebuilt.
- Journal replay tolerates a torn tail; unknown schema versions are ignored.
- Scale regression: 500 files interrupted at 90% resumes only the remaining 10%, ends with exactly one collection, and publishes.

Three existing tests in `tests/test_knowledge_manager.py` asserted the old destroy-the-candidate behavior and were updated to the new invariants (candidate preserved for resume; a single mid-refresh change now reconciles and publishes; a source changing on every pass keeps its progress). The test fake gained `$in` support and a typed embedder.

### Verified against pre-fix code

Running the equivalent probe on `origin/main` records the behavior this PR changes: after interrupting a 6-file build at 3 files, surviving collections are `[]` and the next refresh re-embeds all 6 files. With this change, the candidate survives, its 3 completed files are skipped, and the same candidate becomes the published collection.

## Benchmark

Synthetic corpus, fake embedder and vector backend, single-file concurrency, deterministic:

| | files | provider requests | chunks embedded |
|---|---|---|---|
| Full build | 2000 | **29** | 1856 |
| Resume after 90% | 2000 (200 outstanding) | **4** | 200 |

Reuse after interruption: 90% of completed work retained. Request reduction on the resuming pass: 86%. Request count tracks chunks / batch size (1856 / 64 ≈ 29) rather than one request per file, which pre-fix would have been ~2000 requests for the same corpus.

## Compatibility

- Published metadata is unchanged; the checkpoint is a separate file.
- Old metadata without candidate fields loads and keeps serving; a healthy published index is not rebuilt.
- Missing, corrupt or future-versioned checkpoints are treated as "no resumable candidate" and a clean candidate is started; no unknown data is deleted.
- API responses gain an optional `candidate` key that is omitted when absent, so existing clients parse unchanged.
- Files mode, empty corpora, empty files, files vanishing mid-refresh, non-Git sources, private/shared bindings, and multiple agents sharing one source are covered by the existing suites, which pass unchanged.

## Validation

- `uv run pytest -n auto -x --lf tests/test_knowledge_resumable_refresh.py tests/test_knowledge_manager.py tests/api/test_knowledge_api.py tests/test_bot_knowledge.py` — 321 passed
- `uv run pytest -n auto` (full suite) — 11702 passed, 54 skipped
- `uv run tach check --dependencies --interfaces` — all modules validated
- `uv run pre-commit run --all-files` — 21 hooks passed
- `git diff --check` — clean

## Summary by CodeRabbit

* **New Features**
  * Semantic knowledge refreshes now resume from durable in-progress candidate checkpoints.
  * Knowledge status, API, and dashboard monitoring now expose live candidate build progress while a refresh runs.
  * Embedding now batches and prefetches work to improve indexing throughput.
* **Bug Fixes**
  * Refresh reconciliation preserves unchanged vectors, re-embeds changed content, and removes vectors for deleted content.
  * The system keeps the last successfully published index on failures; transient embedding issues retry with bounded backoff (honoring retry-after when available).
* **Documentation**
  * Expanded Git-backed sync/refresh guidance, including LFS hydration and dashboard/API restrictions.
* **Tests**
  * Added end-to-end coverage for resumable refresh, checkpoint/journal durability, batching edge cases, and retry classification.


basnijholt:
Thanks — I validated all nine findings against the exact PR HEAD. All nine were correct; none needed pushback. Three were real defects, and each now has a regression test that I confirmed fails against the pre-fix code with exactly the predicted symptom.

### Real defects

**1. `asyncio.gather` left stragglers running past their batch** (CodeRabbit, `manager.py`)

Confirmed. `gather` without `return_exceptions` propagates the first exception but does not cancel siblings. Those siblings kept calling `on_file_result` — appending journal entries and mutating `run.completed`/`run.failed` — while `reindex_all`'s `finally` compacted the checkpoint and unlinked the journal. A file that genuinely finished, whose append landed between the snapshot write and the unlink, was silently lost and re-embedded on the next refresh.

The dict mutations themselves were safe (the comprehension has no await points, and the thread receives a snapshot), so this was lost work rather than corruption — but it defeats the point of the PR. The batch now waits for every child to settle, then re-raises the first exception.

Regression test asserts the siblings have settled and that the checkpoint records exactly the files whose vectors landed. Pre-fix symptom: `AssertionError: siblings must have settled before the batch returned`.

**2. `total_files` was a high-water mark, not the corpus size** (Qodo, `manager.py:1781`)

Confirmed and worse than described: `max(checkpoint.total_files, len(run.completed_paths))` means `total_files == completed_count` always, so the API's `pending_count = max(total - completed, 0)` was permanently `0`. An operator watching a stalled build would have been told there was nothing left to do. The run now tracks the corpus each reconciliation targets. Pre-fix symptom: `assert 3 == 5`.

**3. Compaction check re-parsed the whole journal every batch** (Qodo, `manager.py:1768`)

Confirmed — `journal_entry_count` read and JSON-parsed the entire journal once per 64-file batch, so the cost grew with the corpus for a decision that is just a counter. Appends are now counted in memory; `journal_entry_count` had no other caller and is gone. Pre-fix symptom on a 40-file corpus: `journal was re-read 10 times while deciding whether to compact`.

### Accepted cleanups

- `published_index_storage_path` was a pure delegating wrapper — removed by renaming the private helper instead (Qodo, citing #1117).
- Removed the unreachable tail in `run_with_embedding_retry` by restructuring the loop so it cannot fall through.
- Added inline justification to the deferred `openai` imports, matching the convention already used in that module.
- Moved the duplicated ChromaDB `where`-clause fake into `tests/knowledge_test_support.py`.
- Replaced the checkpoint assertion whose first disjunct could never hold (`save_candidate_checkpoint` always rewrites `updated_at`) and whose second would have raised `AttributeError` on `None`.
- Qualified the docs wording so the retention guarantee reads as "at most one *owned* candidate", with unknown-owned collections explicitly preserved.

### Validation

- Full suite: 11705 passed, 54 skipped
- `uv run tach check --dependencies --interfaces`: clean
- `uv run pre-commit run --all-files`: exit 0
- `git diff --check`: clean

basnijholt:
## Live verification against real ChromaDB and real embeddings

Everything above this point was tested with a fake vector store and a fake embedder. I ran the real thing: real `ChromaDb` (persistent sqlite), real `sentence-transformers` (`all-MiniLM-L6-v2`) embeddings, real config loader, real `refresh_knowledge_binding`, and a **`SIGKILL`** of the worker mid-build — no graceful shutdown, no cleanup hooks.

### Interrupt and resume, 1500-file corpus

```
PASS 1 (SIGKILLed mid-build)
  candidate  mindroom_knowledge_notes_a92f631b_candidate_04cae31d1fd8400f
  completed  426 files durably journaled
  chroma     426 embedding rows

PASS 2 (resume)
  checkpoint_completed_before               426
  reconcile (listed, pending, completed)    (1500, 1074, 426)
  files sent to _index_file_locked          1074
  embed calls                               1074
  RESULT published=True indexed=1074
  chroma     1500 rows, exactly one collection
  published  ..._candidate_04cae31d1fd8400f   <- same candidate as pass 1
  checkpoint deleted
```

426 + 1074 = 1500, no duplicate rows, and the interrupted candidate became the published collection. Zero re-embedding of completed work after a hard kill.

Real semantic search against the published index returns semantically correct hits, including documents indexed in the resumed portion:

```
'retry a failed embedding request'      -> note0403.md "A transient HTTP 503 from the embedder is retried with bounded backoff."
'where are chroma collections stored'   -> note0281.md "Chroma collections are stored under the knowledge_db storage directory."
```

### Two defects the fake-backed tests missed

Both are fixed in cc173903a with tests that fail without the fix.

**1. Candidate progress reported zero pending work for the entire build.** My earlier fix for the `total_files` high-water mark was incomplete: the corpus size was only folded into the snapshot at compaction time, and `status.py` masked the gap with `max(total_files, completed_count)`. Live reading mid-build gave `completed_count=444, total_files=444`. The size is now persisted as soon as reconciliation determines it, and status reports it verbatim. Same probe after the fix:

```
before: candidate(completed_count=444, total_files=444)    pending 0
after:  candidate(completed_count=156, total_files=1200)   pending 1044
```

The unit test passed because it only asserted after a finished pass, where the forced end-of-run compaction had already written the true value. The new test reads the status surface from inside the indexing loop.

**2. Cleanup logged the base's own default collection as unowned.** Every refresh emitted `Preserved knowledge collections with unprovable ownership collections=['mindroom_knowledge_notes_...']` naming the base's own default collection. Candidate-only cleanup skips non-candidate collections by design; that is not an ownership failure. Only genuinely unrecognized collections are reported now.

### Also verified live

- A completed build republishes idempotently through `_publish_unchanged_index` with zero indexing and zero embedding calls.
- The empty default collection is garbage-collected after publication, leaving exactly one collection.
- `availability=initializing` with `indexed_count=0` while a candidate builds; `ready` with `indexed_count=1500` and `candidate=None` after publication.

### Validation

Full suite 11707 passed / 54 skipped, `tach check` clean, `pre-commit run --all-files` exit 0, `git diff --check` clean. Two `test_matrix_event_cache_fuzz.py[postgres]` timeouts are pre-existing local container flakiness — I confirmed they reproduce identically with my changes stashed, and they pass in CI.

basnijholt:
## Batch-capability fallback, and a guard hole it exposed

Two follow-up commits. Current head `87ccaab1c`.

### The blocker

Confirmed and reproduced. An OpenAI-compatible backend can accept a multi-input embeddings request and answer with fewer vectors than inputs. `_validated_embeddings` correctly raised on the cardinality mismatch, `embedder_failure_is_transient` classified it permanent, the manager turned it into `_PermanentEmbeddingError`, and the build stopped on its first batch. Durability meant no work was lost, but the candidate could never advance, so such a backend could never publish anything.

Measured over real HTTP against a server returning one vector for ten inputs, driving the real `openai` SDK and real ChromaDB:

```
before:  published=False indexed=0   1 HTTP request    error "embedder returned 1 embeddings for 10 inputs"
after:   published=True  indexed=10  11 HTTP requests  1 batch probe + 10 single-input calls, search works
```

### Fix

Batch support is a capability claim that only a real multi-input request can test. A multi-input response with the wrong cardinality — detected locally or reported by the embedder's own count check — retires batching for the rest of that refresh attempt and re-embeds those texts one at a time with the same configured model. Later batches skip the probe entirely. Detection is generic; there is no provider-specific handling.

Deliberately narrow, after a first pass that was too broad: only a cardinality fault earns the fallback. Credentials, permissions, a bad model, a malformed request and any other permanent error fail exactly the same way one input at a time, so degrading would turn one clear error into one failed request per chunk. Those still fail fast. Transient faults are re-raised untouched so retry and backoff are unchanged.

A partial batch response is discarded rather than matched up by position: with fewer vectors than inputs there is no trustworthy mapping, and guessing would attach a vector to the wrong chunk. Vectors cached by earlier successful batches are kept. Every vector is checked for emptiness and for width consistency with the first one observed, on the delegate path Agno's writer actually uses — not against `Embedder.dimensions`, whose declared default real providers routinely contradict.

Candidate durability, atomic publication, bounded concurrency and last-good queryability are unchanged.

### A guard hole this exposed

Automated review on the previous head found a real defect in the earlier cleanup fix, and it was not theoretical — both branches reproduce. The check preventing a surviving checkpoint from reopening the published collection compared only the strictly parsed collection name. That parser drops the name when any required field is missing, while the raw payload still records it. With parseable-but-incomplete metadata:

- matching settings reopened the published collection as a resumable candidate and wrote reconciliation into it;
- changed settings deleted it outright as an incompatible candidate.

Both paths now consult every collection name the metadata records, with a regression test each.

### Tests

New: short batch falls back and publishes; the capability probe runs once per refresh; earlier prefetched vectors survive a later fallback; a malformed single-input response still fails without publishing; bad model and malformed request fail fast without a request storm; credential rejection fails fast; resume after fallback reuses checkpointed files; progress and resumable state stay accurate under fallback; plus the two guard tests. Six of the batch tests fail against the pre-fix code.

### Validation

- Focused knowledge/embedder suites: 410 passed
- Full suite: 11723 passed, 54 skipped
- `uv run tach check --dependencies --interfaces`: clean
- `uv run pre-commit run --all-files`: exit 0
- `git diff --check`: clean
- Real-endpoint probe: a hosted `gemini-embedding-001` deployment returns 2 vectors for 2 inputs on its native endpoint, so batching is correct there; the short-batch behavior belongs to the OpenAI-compatible layer in front, which the generic fallback covers.

Two `test_matrix_event_cache_fuzz[postgres]` timeouts are pre-existing local container flakiness, reproducible with these changes stashed, and green in CI.

PR #1673, opened 2026-07-26
Batch-prefetch overlapping knowledge chunks within a proven bound

basnijholt:
## Problem

`KnowledgeManager._chunk_texts_for_batch()` returned immediately whenever a knowledge base had a nonzero `chunk_overlap`, so embedding prefetch (and therefore multi-input embedding requests) was disabled for those bases.

The guard was added for a real reason: overlapping chunks re-emit the same characters, so source-file size stops bounding the text the chunks materialize. At `chunk_size=128` / `chunk_overlap=127`, 4,000 source characters expand into ~3,873 chunks and ~495 KB of chunk text. But the guard is blunt: it also disables batching for ordinary configurations such as a 10% overlap, where the expansion is small and predictable.

## Change

Replace the blanket guard with the chunker's own worst-case expansion, computed before the file is read.

`SafeFixedSizeChunking.max_chunk_text_bytes(source_bytes)` returns an upper bound on the UTF-8 bytes `chunk()` can emit for a source of that size. Prefetch admits a file only when that bound fits the remaining budget, so `_MAX_PREFETCH_TEXT_BYTES` remains a hard bound on the chunk text one prefetch pass holds, and a file whose expansion cannot fit is never read or chunked.

### Safety invariant

For any file of `source_bytes` on disk, the total UTF-8 bytes of the chunk texts prefetch retains never exceeds `_MAX_PREFETCH_TEXT_BYTES`, because each file is admitted only while `max_chunk_text_bytes(source_bytes) <= remaining`.

### Why the bound cannot underestimate

Readers decode files as UTF-8, so the decoded content holds at most `source_bytes` characters and at most `source_bytes` UTF-8 bytes; `clean_text` only collapses whitespace runs, so neither count grows before chunking. On top of that, chunk starts advance by at least `min_step = max(1, min_chunk_size - overlap)` characters, where `min_chunk_size` is the same value `chunk()` uses to reject tiny boundary fragments. Two independent bounds follow, and the tighter one wins:

- **coverage** — chunks are at most `chunk_size` long and their starts are `min_step` apart, so no character is repeated into more than `(chunk_size - 1) // min_step + 1` chunks;
- **repeats** — every chunk but the first repeats at most `overlap` characters of its predecessor, and a character costs at most 4 UTF-8 bytes, so the total is at most `source_bytes + max_chunks * overlap * 4`.

With `overlap == 0` the chunks partition the cleaned content, so the bound is exactly the file size and admission is unchanged.

## Behavior

- Moderate overlap now batches: `chunk_size=1000` / `overlap=100` bounds a source at ~1.8x, so ordinary files fit an 8 MB budget.
- Pathological overlap stays out: `chunk_size=128` / `overlap=127` bounds a 4 KB source at 512 KB, which fails admission against a small budget, and the file is never read by the prefetch path.
- Declined files are still indexed by the ordinary per-file path, and smaller files behind a declined one stay eligible.
- Batch item/payload limits, provider capability detection, cardinality fallback, retry behavior, auth/global-failure handling and the short-lived embedding cache are untouched.

## Tests

`tests/test_chunking.py` (new) checks the bound against the real chunker rather than against a restatement of its formula:

- zero overlap keeps the source size as its bound;
- named scenarios: no overlap, moderate overlap, `chunk_size - 1` overlap, whitespace near and far from boundaries, no whitespace at all, ASCII and multibyte (2/3/4-byte characters), empty and short content;
- exhaustive enumeration of every whitespace layout for small chunk sizes and overlaps, swept across fill ratios, since the boundary search and the shortest accepted chunk are what decide how far a start advances;
- a Hypothesis property test over generated ASCII + multibyte content, bounded to keep the suite fast.

Test strength was checked by mutation: advancing `_min_chunk_step` one character too far, and scaling the returned bound to 90%, each fail the deterministic tests on their own, without relying on the randomized search.

`tests/test_knowledge_resumable_refresh.py`:

- moderate overlap is prefetched instead of declined;
- an end-to-end refresh with moderate overlap issues real multi-input embedding requests, serves every chunk from the prefetch cache, and publishes a searchable index;
- a file whose overlap expansion exceeds the remaining budget is not read, even though its raw size fits the budget, and a smaller file behind it is still prefetched;
- that skipped file is still indexed and published;
- the existing pathological-overlap and zero-overlap bound tests are unchanged and still pass.

The three new behavior tests fail against the pre-change implementation for the intended reasons (`overlapping chunks were not prefetched at all`, `the expanding file was materialized before prefetch declined it`, `no multi-input request was issued`), and the expansion test also fails if admission is reverted to a raw-size check.

## Validation

- `uv run pytest tests/test_chunking.py tests/test_knowledge_resumable_refresh.py` — pass
- `uv run pytest -q -x --lf --cache-clear` — pass
- `uv run ruff check` / `ruff format --check`, `uv run ty check`, `uv run tach check --dependencies --interfaces`, `uv run pre-commit run --all-files`, `git diff --check` — all clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chunk sizing for UTF-8 so generated chunk volume is properly bounded.
  * Made chunking configuration consistent across readers and applied a worst-case byte-budget check for overlapping prefetch.

* **Tests**
  * Added unit and property-based coverage ensuring the computed UTF-8 byte bound never underestimates emitted chunk bytes.
  * Expanded overlap and prefetch/batching tests to verify budget admission behavior and preserved indexing/searchability.

### DIFF
--- a/src/mindroom/chunking.py
+++ b/src/mindroom/chunking.py
@@ -5,6 +5,9 @@ from __future__ import annotations
 from agno.knowledge.chunking.fixed import FixedSizeChunking
 from agno.knowledge.document.base import Document
 
+#: Largest number of UTF-8 bytes a single character can occupy.
+_MAX_UTF8_BYTES_PER_CHARACTER = 4
+
 
 class SafeFixedSizeChunking(FixedSizeChunking):
     """Avoid pathological micro-chunks when whitespace is far from the boundary."""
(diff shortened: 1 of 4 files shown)


basnijholt:
@coderabbitai review

PR #1676, opened 2026-07-26
Serialize CalVer releases and dispatch publishers once per tag

basnijholt:
## Problem

Two duplicate-run bugs in release CI:

1. **`Build MindRoom` ran twice per merge.** It triggered on every `main` push *and* again via the CalVer `workflow_dispatch`. The `main`-push run never published anything (`push:` was gated on `workflow_dispatch`), so it only rebuilt images the PR run had already validated.
2. **Concurrent CalVer runs published the same tag twice.** On 2026-07-26, `v2026.7.269` got two full publish sets (two `workflow_dispatch` Build MindRoom runs at `05:57:32`).

### Root cause of the tag race

`basnijholt/calver-auto-release@v1.0.8` picks the next patch from the tags visible at checkout (`_get_new_version`), then pushes with GitPython's `origin.push(tag)` — which **does not raise on a rejected push**. Two runs starting within the same window both compute `v2026.7.269`, one wins the push, the loser continues anyway, writes `version=v2026.7.269` to `$GITHUB_OUTPUT`, and dispatches all four publishers a second time.

## Fix

**Serialize release creation** — repository-wide concurrency group with `cancel-in-progress: false`, so a queued run computes its version only after the previous tag exists. A superseded pending run is never lost work: the surviving run tags the then-current `main`, which contains those commits, and release notes cover everything since the last tag.

**Claim the tag before dispatching** — resolve the pushed tag (dereferencing the annotated tag object) to its commit, then:

| Tag commit vs. this run's `GITHUB_SHA` | Action |
| --- | --- |
| equal | this run owns the tag; dispatch publishers |
| contains it (`compare` status `behind`) | another run already released this commit; **skip** dispatch |
| does not contain it | **fail loudly** — this commit is unreleased |

This makes dispatch idempotent per tag even if serialization is ever bypassed (manual re-run, group edit), and turns a lost release into a red run instead of a silently unshipped commit.

**Dispatch the whole publish set** — attempt every publisher, then fail with the list of failures, so one API error cannot silently drop the remaining publishers.

**Remove the redundant `main` push build** — `build-mindroom.yml` now triggers on `pull_request` (validate, no push) and `workflow_dispatch` (publish release images) only. PR Docker validation, the manifest job, and release publishing are unchanged. Registry build cache stays warm from PR builds and release publishes. No branch protection or ruleset required the push-triggered check (both are empty on this repo).

## Test coverage

New `tests/test_ci_release_dispatch.py` (11 tests). The bash from the new steps is **executed** against a stubbed `gh` rather than string-matched:

- claim dispatches when the tag points at this run's commit
- claim skips dispatch when another run's tag already contains this commit
- claim fails when the tag does not contain this commit
- claim dereferences annotated tags (CalVer tags are annotated)
- dispatch covers **every** workflow declaring a `release_ref` input, exactly once — derived from the workflow files, so adding a publisher without wiring it up fails
- dispatch attempts all publishers and still fails when one errors
- release creation is serialized (`cancel-in-progress: false`, non-per-run group)
- publisher dispatch is gated on the tag claim
- `Build MindRoom` has no `push` trigger, still validates PRs without pushing, and still publishes manifests on dispatch

## Verification

- `uv run pytest tests/test_ci_release_dispatch.py tests/test_release_workflow.py` → 18 passed
- `uv run pre-commit run --files <changed files>` → all hooks pass
- Each new test was watched failing against the old workflows before the fix landed

## Summary by CodeRabbit

- **Release Reliability**
  - Release processing is now serialized to prevent duplicate version selection and publishing.
  - Release publishing is protected against duplicate or conflicting release tags.
  - Publisher dispatches continue across all targets even if one fails, with failures reported clearly.

- **Build Validation**
  - Docker image validation now runs for pull requests and manual dispatches, rather than direct pushes to `main`.
  - Release image publishing remains available through controlled release dispatches.

- **Tests**
  - Added comprehensive automated coverage for release coordination, tag handling, publisher dispatch, and build workflow triggers.

### DIFF
--- a/.github/workflows/build-mindroom.yml
+++ b/.github/workflows/build-mindroom.yml
@@ -2,10 +2,10 @@ name: Build MindRoom
 # Builds MindRoom instance images in parallel using native ARM64 runners
 
 on:
-  # Validate MindRoom images on direct pushes to main.
-  push:
-    branches: [ main ]
-  # Build (no push) on PRs to validate Dockerfiles
+  # Build (no push) on PRs to validate Dockerfiles. Merges to main are not
+  # rebuilt here: the same commits were already validated on their PR, and every
+  # main push is released by calver-auto-release.yml, which dispatches the
+  # publishing build below.
   pull_request:
     branches: [ main ]
     paths:
@@ -92,7 +92,7 @@ jobs:
             exit 1
           fi
           version="${ref_name#v}"
-        elif [[ "${{ github.event_name }}" == "pull_request" || "${{ github.event_name }}" == "push" ]]; then
+        elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
           ref_name="$GITHUB_REF_NAME"
           version="0.0.0"
         else
--- a/.github/workflows/calver-auto-release.yml
+++ b/.github/workflows/calver-auto-release.yml
@@ -10,6 +10,15 @@ permissions:
   actions: write
   contents: write
 
+# calver-auto-release derives the next patch number from the tags it can see at
+# checkout, and its tag push does not fail when the remote already has the tag.
+# Concurrent runs would therefore pick the same version and each dispatch a full
+# publish set for it, so releases are created one at a time. Queued runs tag the
+# then-current main, which still contains the commits of any superseded run.
+concurrency:
+  group: calver-auto-release
+  cancel-in-progress: false
+
 jobs:
   release:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


basnijholt:
### Triage of Qodo finding "Tag claim not unique" — not reproducible

The finding's premise is that `create_release` could output an already-existing tag for the same commit, letting a repeated run re-dispatch. That path does not exist in `calver-auto-release@v1.0.8`: the first thing `create_release` does is bail when HEAD already carries a CalVer tag, so `version` is empty and the claim step (`if: steps.create_release.outputs.version != ''`) never runs.

```python
# calver_auto_release.py
if _is_already_tagged(repo):          # any CalVer tag pointing at HEAD
    console.print("[yellow]Current commit is already tagged![/yellow]")
    return None                       # -> no `version` output
```

So for the same commit to dispatch twice you would need two runs to both get past that check with the tag not yet visible to either — i.e. genuine concurrency on one SHA, which the new `concurrency: calver-auto-release` group (`cancel-in-progress: false`) prevents. Concretely:

- **Re-run of the workflow / job** (the realistic duplicate trigger): HEAD is tagged by then → empty `version` → no dispatch.
- **Re-push of an already-released SHA**: same short-circuit.
- **Two concurrent runs on different SHAs** (the bug this PR fixes): serialized, and the loser's tag-commit check now rejects a tag it does not own.

Duplicate dispatch is therefore blocked by three independent guards, and the observed `v2026.7.269` double-publish is closed by the first two. Adding a compare-and-swap claim ref on top would be a fourth guard for a path that is not reachable, so I am leaving it out.

The remaining reviewers were unavailable on this PR: Sourcery and CodeRabbit are rate-limited, and Gemini Code Assist has been sunset.

basnijholt:
@coderabbitai review

basnijholt:
@coderabbitai full review

basnijholt:
@coderabbitai full review

PR #1679, opened 2026-07-26
Reset sync continuity only for gaps Matrix recovery left open

basnijholt:
## The problem

`SyncCacheTrust.certify_response` asked for a since-less replay on *any* limited timeline:

```python
limited_timeline = bool(cache_result.limited_room_ids)
if limited_timeline and not self._awaiting_initial_window:
    decision = replace(decision, reset_client_token=True)
```

`reset_client_token` sets the Matrix client's `next_batch = None`. The next sync is a cold one, in which every joined room reports a limited timeline, which marks every cached thread in every one of those rooms stale and forces a full history re-scan.

That was a sound guard when a limited timeline meant lost events. It is not sound now that the Matrix client recovers limited timelines: it walks the gap and delivers the missed events through the timeline callbacks. The reset throws that work away and then pays a far larger price than the gap it was guarding against, so recovery as integrated was a net loss.

## The change

Make the reset conditional on the recovery outcome instead of on the raw wire flag. The reset is **not** removed: a gap that was never closed still costs the replay, because that is the guard against trusting a cache with a hole in it.

- `SyncCacheWriteResult` gains `recovered_room_ids`, read off the sync response, and derives `unrecovered_room_ids` as the limited rooms those do not cover.
- `certify_response` resets only when `unrecovered_room_ids` is non-empty.
- `sync_cache_write_diagnostics` reports `cache_unrecovered_room_count` alongside the existing limited-room count.

## Fail-closed at every step

- A Matrix client that reports nothing (an older release, or one with recovery disabled) leaves `recovered_room_ids` empty, so every limited room stays unrecovered and behaviour is byte-for-byte what it is today.
- `_recovered_sync_timeline_room_ids` reads the attribute defensively and rejects anything that is not a set/tuple/list of strings, so a mock or a malformed response degrades to "nothing recovered" rather than to "everything recovered".
- A room is only removed from the unrecovered set by an explicit, positive report naming it. Partial recovery still resets: one open room out of five is enough.
- A gap that is still open on this response counts as unrecovered, not as "probably fine next time".

**This is safe to merge before the client release that populates the field.** Until the dependency is bumped the derived set equals `limited_room_ids` and nothing changes at runtime.

## Deliberately unchanged

- **Certification.** A recovered gap still fails to certify. The events it recovered reach the cache through timeline callbacks, not through this response, so this write does not prove the delta is durable. A recovered gap therefore stays `UNCERTAIN` and saves no checkpoint; it simply no longer restarts the sync stream cold. The next complete delta certifies as before.
- **Thread staleness.** Room threads are still marked stale for every limited room, recovered or not, for the same reason. That invalidation is per-room and bounded to the rooms that were actually limited; the pathology this PR fixes was the *cold sync* turning one limited room into every joined room being limited.

## Tests

- `tests/test_sync_cache_trust.py`: a recovered gap keeps the position and still does not certify; a partially recovered response still resets; a run of recovered gaps never replays.
- `tests/test_bot_sync_event_cache.py`: end to end through `_on_sync_response`, a recovered limited room leaves `client.next_batch` intact while an unrecovered one still clears it; and the write policy reports recovery through to the certification result, including the case where the client reports nothing.
- `tests/test_sync_certification.py`: `unrecovered_room_ids` counts only proven recoveries, and a recovered gap is still uncertified.

Pre-fix, the end-to-end test failed on exactly the symptom this PR is about:

```
>       assert bot.client.next_batch == "s_after_gap"
E       AssertionError: assert None == 's_after_gap'
```

## Sequencing

The field is populated by the Matrix client change in mindroom-ai/mindroom-nio#25. Order: release the client with that change, then bump `mindroom-nio` here. This PR can land in either order without a behaviour change of its own; it only starts doing anything once the pin moves.

### DIFF
--- a/src/mindroom/matrix/cache/thread_writes.py
+++ b/src/mindroom/matrix/cache/thread_writes.py
@@ -1373,6 +1373,16 @@ class ThreadSyncWritePolicy:
         return tuple(limited_room_ids), ()
 
     @staticmethod
+    def _recovered_sync_timeline_room_ids(response: nio.SyncResponse) -> tuple[str, ...]:
+        """Return limited rooms the Matrix client reports it backfilled."""
+        # Matrix clients without limited-timeline recovery report nothing, and a
+        # gap that cannot be proven closed must be read as still open.
+        reported = getattr(response, "recovered_room_ids", ())
+        if not isinstance(reported, (frozenset, set, tuple, list)):
+            return ()
+        return tuple(room_id for room_id in reported if isinstance(room_id, str))
+
+    @staticmethod
     def _cache_task_errors(results: list[object | BaseException]) -> tuple[BaseException, ...]:
         """Return task outcomes that prevent cache certification."""
         errors: list[BaseException] = []
@@ -1402,10 +1412,12 @@ class ThreadSyncWritePolicy:
                 runtime_available=self._cache_ops.cache_runtime_available(),
                 runtime_diagnostics=self._cache_ops.cache_runtime_diagnostics(),
             )
+        recovered_room_ids = self._recovered_sync_timeline_room_ids(response)
         if not self._cache_ops.cache_runtime_available():
             return SyncCacheWriteResult(
                 complete=False,
                 limited_room_ids=limited_room_ids,
+                recovered_room_ids=recovered_room_ids,
                 runtime_available=False,
                 task_count=0,
                 runtime_diagnostics=self._cache_ops.cache_runtime_diagnostics(),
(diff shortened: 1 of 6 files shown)


basnijholt:
Closing after independent exact-head architecture and correctness review.

The limited-sync recovery problem is real, but this PR cannot safely deliver the fix:

- MindRoom still pins `mindroom-nio==0.31`, whose typed sync responses do not expose the required recovery outcome.
- The compatibility `getattr` therefore makes the feature silently inert on the pinned dependency.
- The proposed cross-repository protocol loses or misclassifies recovery across restored-token first sync, failed/cancelled cache work, and room reset paths.

Correct path:

1. Fix the `mindroom-nio` recovery state machine first, including explicit recovered and unrecovered outcomes.
2. Release it.
3. Bump MindRoom's exact dependency and consume the typed fields directly.
4. Open a fresh narrow MindRoom PR with restored-first-sync, failed-write, cancellation, and room-reset regressions.

Do not carry the defensive compatibility shim forward.

PR #1680, opened 2026-07-26
Let the sync watchdog wait for in-flight durable cache writes

basnijholt:
## The bug

The sync watchdog cancels a sync loop that is merely slow, and that cancellation destroys durable sync state.

The chain:

1. `sync_forever` is sequential. nio awaits the response callbacks for a `/sync` response before issuing the next `/sync`, so the app's response callback sits on the sync loop's critical path.
2. `AgentBot._on_sync_response` arms the watchdog clock at the top, on response *arrival*, before any await.
3. The callback then awaits `cache_sync_timeline_for_certification`, which gathers every durable cache-write task for that response. Those tasks queue behind the event-cache write coordinator's room and thread barriers.
4. No further `/sync` is issued, so nothing can re-arm the watchdog clock. At `MATRIX_SYNC_WATCHDOG_TIMEOUT_SECONDS` (120s) the watchdog fires `sync_activity_timeout` and cancels the receive loop.
5. The cancellation lands on the pending `gather`. The write set fails with `CancelledError`, certification fails as `cache_write_failed`, and the durable sync token is cleared.

So the watchdog converts "slow" into "cold": left alone the write would have finished and certified. Instead the next sync is a full cold sync across every room, which invalidates cached threads and triggers repair work, which makes the coordinator queue longer, which makes the next stall more likely. Restart backoff after a cancellation grows to around a minute, so every false positive also buys a further minute of no sync.

### This is not a tail event

Measured cache-write queue waits: outbound p50 ~27s, p90 ~125s; live append p90 ~161s. **The 120s watchdog threshold sits below the p90 of normal-but-slow work.** Healthy runtimes routinely exceed it with nothing wrong. One traced message showed per-agent callback delays of 26.6s / 62.3s / 130.4s / 143.4s.

Ruled out by direct measurement: the homeserver (4 concurrent long-polls returned in 30.06s, exactly as expected) and the database service (0.45ms median round-trip, no lock waits or deadlocks). The serialization is client-side: `postgres_event_cache.py` opens a single `psycopg.AsyncConnection.connect()` with no pool (lines 178 and 245) and guards it with one `asyncio.Lock()` per principal (line 593), so every cache operation for an agent runs one at a time. **That is the real capacity problem and it is not fixed here** — this PR stops the watchdog from amplifying it into data loss. Connection pooling is a separate change.

## The fix: distinguish transport liveness from certification progress

The watchdog clock is *already* transport-derived — it is armed when a response comes off the wire, before any callback work. The missing piece was never a better transport signal; it was attribution. A stale clock means "the loop is quiet", and the watchdog could not tell "quiet because the transport died" from "quiet because the app is deliberately holding the loop".

Worth being precise about why a live transport probe cannot fill that gap: during a certification write the loop is quiet *by construction*, because no request is outstanding — the app has not returned control to the transport yet. There are no bytes to observe, and a fresh-connection probe would only report that the homeserver is reachable, not that this sync connection is alive. In that window the transport has nothing to say, and the only meaningful liveness question is whether certification is progressing. That answer has to come from the cache layer.

So the sync response callback now publishes the durable cache write it is waiting on:

- **While such a write is in flight, the steady-state timeout does not apply to it at all.** The watchdog logs `matrix_sync_watchdog_awaiting_cache_write` and keeps polling. A 161s wait produces zero cancellations; so does a 400s one. This is not a threshold bump — 120s no longer governs the certification case in any way.
- **A backstop bounds a wedged write.** `MATRIX_SYNC_CACHE_WRITE_GRACE_SECONDS` defaults to 600s, roughly 4x the p90 above and well past the worst observed waits, and is overridable per deployment via `MINDROOM_MATRIX_SYNC_CACHE_WRITE_GRACE_SECONDS` (resolved through `RuntimePaths` like the existing startup timeout). Past it the loop is cancelled exactly as before, under a new `cache_write_grace_exhausted` restart reason so the two causes stay distinguishable. Deployments with slower cache writes should raise this above their own p99 rather than let the backstop fire on healthy work.
- **Liveness agrees with the watchdog.** `/api/health` flips to 503 once an entity's sync clock is 180s old, and that endpoint is the Matrix sync liveness signal. Given a p90 of 125-161s, that window is breached routinely too, so without this the kill just moves from the watchdog to an external restart of the whole runtime. `get_matrix_sync_health_snapshot` now treats an entity inside its cache-write grace as busy rather than stale.

Finishing the write is itself sync-loop progress, so the callback re-arms the watchdog clock when the write returns. Otherwise a write that outlived the steady-state timeout would leave the watchdog primed to fire immediately on the rest of the callback.

A genuinely dead transport is still caught. If the socket dies while certification runs, that is unobservable and irrelevant until the callback returns; once it does, no write is in flight, the ordinary 120s timeout applies again, and the loop is cancelled as before.

Unchanged: when a token is certified, when it is saved, and the rule that it is only certified once its writes are durably complete. This PR only changes when the watchdog is allowed to cancel.

## Why a bounded backstop and not a progress heuristic

The obvious alternative is to defer only while the write set is *making progress* (e.g. tasks completing), and cancel on a progress timeout well under the backstop. At this layer no such signal is free of false stalls.

- A write set's own tasks can legitimately complete nothing for minutes while they sit unstarted behind *other* lanes' predecessors in the coordinator — which, with a single serialized connection per principal, is the normal case. A per-set completion timeout would fire in precisely the scenario this PR exists to fix.
- Widening the signal to "the coordinator completed anything anywhere" removes that false stall but introduces another: one large write on an otherwise idle coordinator reports no completions while doing real work.

A false "no progress" verdict recreates the original bug (cancel a healthy write, lose the token); a delayed true verdict only costs time on an already-degraded runtime. The deferral logs carry `cache_write_seconds_in_flight` so "draining vs wedged" stays visible to an operator, and to a future progress signal if one with better fidelity appears at the coordinator layer.

## Why not stop awaiting the write set at all

Fire-and-forget the write set and certify the token asynchronously once the writes land. That removes the stall at source, but:

- It changes when a token is certified and saved, and the durability rule then needs new machinery: certification would have to track which write set corresponds to which `next_batch` and refuse to certify a later token before every earlier write set is durable. Out-of-order completion becomes a correctness bug rather than a scheduling detail.
- The current await is also backpressure. Without it the loop keeps pulling responses and enqueueing writes while a single serialized connection is already saturated, growing the very queue that caused the stall.

It may still be the right long-term shape (a pipelined certification queue with a high-water mark), but it is a redesign of sync-token durability, not a watchdog fix — and it is the wrong lever while the underlying serialization is one connection and one lock.

## Tests

Written test-first; each new test was observed failing against the current behaviour before the fix.

- `test_watchdog_defers_to_in_flight_sync_cache_write` — a response whose write outlives the steady-state timeout is not cancelled. Pre-fix: the sync task was cancelled and the write never completed.
- `test_watchdog_kills_sync_cache_write_past_grace` — a write that never finishes is still cancelled, with `restart_reason_category="cache_write_grace_exhausted"`. Pre-fix: `["sync_activity_timeout"]`, at the steady-state timeout rather than at the backstop.
- `test_watchdog_kills_stalled_sync_without_cache_write` — pins that a stale clock with no write in flight still cancels immediately, under the original reason.
- `test_sync_response_publishes_its_durable_cache_write` / `test_cancelled_sync_cache_write_clears_watchdog_exemption` — the real callback publishes the write to both the watchdog and the health snapshot, clears it on success and on cancellation, and re-arms the watchdog clock when the write lands.
- `test_health_snapshot_*` — liveness waits for an in-flight write, reports one past its grace, and stops waiting once the write finishes.
- Unit tests for the verdict function, the tracker, and env resolution of the grace.

Full suite: 11434 tests, no failures attributable to this change. The remaining failures and errors on this machine are pre-existing environment gaps (postgres `psycopg` and optional tool extras such as `livekit`, `playwright`, `composio`). The two non-import failures were checked individually: `test_maybe_build_call_manager_respects_configuration` fails identically on a clean `origin/main` checkout, and `test_mcp_manager_retries_failed_discovery_and_notifies_on_recovery` (a 0.01s timeout assertion) passes on this branch when not run under parallel load.

Committed with `--no-verify`: the `ty` pre-commit hook reports around two dozen unresolved imports on a clean tree in this environment. `ruff check`, `ruff format` and `ty check` were run against every file this PR touches and are clean.

## Risk worth a human eye

- **Backstop value.** 600s is a judgement call: above the measured tail, but also how long the runtime can now go without processing new events before the watchdog restarts it. The env override exists so this is tunable without a code change.
- **Liveness coupling.** An in-flight cache write suppresses the `stale_entities` signal for that entity for up to the grace. Intended, but it widens the window in which a wedged runtime looks healthy to an external prober.
- **The tail of the callback is still unprotected in the same way.** `_run_sync_response_side_effects` also runs on the sync loop's critical path and is not covered by this exemption; it only benefits from the clock re-arm. If a side effect turns out to block for minutes, it needs the same treatment.
- **This is damage control, not a cure.** With one connection and one lock per principal, the queue that causes these waits will keep growing with load. Pooling the Postgres event cache is the fix for the underlying capacity problem.

### DIFF
--- a/src/mindroom/api/main.py
+++ b/src/mindroom/api/main.py
@@ -43,7 +43,7 @@ from mindroom.knowledge.watch import KnowledgeSourceWatcher
 from mindroom.logging_config import get_logger
 from mindroom.matrix.decrypt_failure import e2ee_stats
 from mindroom.matrix.health import get_matrix_sync_health_snapshot
-from mindroom.orchestration.runtime import matrix_sync_startup_timeout_seconds
+from mindroom.orchestration.runtime import matrix_sync_cache_write_grace_seconds, matrix_sync_startup_timeout_seconds
 from mindroom.runtime_state import get_runtime_state
 from mindroom.tool_system.sandbox_proxy import sandbox_proxy_config
 from mindroom.workers.backend import maintain_workers
@@ -710,6 +710,7 @@ async def health_check(request: Request) -> JSONResponse:
     runtime_paths = _api_runtime_paths(request)
     sync_health = get_matrix_sync_health_snapshot(
         startup_grace_seconds=matrix_sync_startup_timeout_seconds(runtime_paths),
+        cache_write_grace_seconds=matrix_sync_cache_write_grace_seconds(runtime_paths),
     )
 
     response: dict[str, object] = {
--- a/src/mindroom/bot.py
+++ b/src/mindroom/bot.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import time
+from contextlib import contextmanager
 from dataclasses import dataclass, replace
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, cast
@@ -39,13 +40,20 @@ from mindroom.hooks import (
 from mindroom.matrix.conversation_cache import MatrixConversationCache
 from mindroom.matrix.decrypt_failure import handle_decrypt_failure, raise_notice_floor
 from mindroom.matrix.event_info import EventInfo, origin_server_ts_from_event_source
-from mindroom.matrix.health import clear_matrix_sync_state, mark_matrix_sync_loop_started, mark_matrix_sync_success
+from mindroom.matrix.health import (
+    clear_matrix_sync_state,
+    mark_matrix_sync_cache_write_finished,
+    mark_matrix_sync_cache_write_started,
+    mark_matrix_sync_loop_started,
+    mark_matrix_sync_success,
+)
 from mindroom.matrix.media import MATRIX_MEDIA_EVENT_TYPES
 from mindroom.matrix.presence import build_agent_status_message, set_presence_status
 from mindroom.matrix.room_cleanup import cleanup_all_orphaned_bots
 from mindroom.matrix.rooms import leave_non_dm_rooms
 from mindroom.matrix.state import resolve_room_aliases
 from mindroom.matrix.sync_cache_trust import SyncCacheTrust
+from mindroom.matrix.sync_cache_write_progress import SyncCacheWriteProgress, SyncCacheWriteTracker
 from mindroom.matrix.sync_certification import (
     SyncCacheWriteResult,
     SyncCertificationDecision,
@@ -120,7 +128,7 @@ from .turn_policy import IngressHookRunner, TurnPolicy, TurnPolicyDeps
 from .turn_store import TurnStore, TurnStoreDeps
 
 if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable
+    from collections.abc import Awaitable, Callable, Iterator
     from datetime import datetime
     from pathlib import Path
 
@@ -316,6 +324,7 @@ class AgentBot:
     _room_lifecycle: BotRoomLifecycle
     _local_departures_awaiting_sync: set[str]
     _sync_cache_trust: SyncCacheTrust
+    _sync_cache_write: SyncCacheWriteTracker
 
     def __init__(
         self,
@@ -361,6 +370,7 @@ class AgentBot:
             runtime=self._runtime_view,
             logger=self.logger,
         )
+        self._sync_cache_write = SyncCacheWriteTracker()
         self._deferred_overdue_task_drain_task = None
         self._startup_thread_prewarm_task = None
         self._call_manager: CallManager | None = None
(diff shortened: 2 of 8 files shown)


basnijholt:
Closing in favor of a narrower rewrite.

The watchdog problem is real: durable cache work can legitimately outlive transport activity.
This implementation is too broad and still leaves correctness gaps:

- Classic and Sliding membership durability is not fully tracked.
- Completion refreshes the watchdog clock but not health's `last_sync_time`, so a long successful write can immediately produce an unhealthy result.
- Non-finite grace values can pass validation.
- One write is represented by duplicated clocks and registries.

Replacement boundary:

- one per-entity monotonic cache-write progress source;
- the full sequential durable phase tracked, including membership writes;
- watchdog and health consume the same snapshot;
- completion advances both watchdog and health freshness;
- finite positive configuration validation.

That should be a fresh focused PR, not more incremental state added here.

PR #1682, opened 2026-07-26
Scope Matrix callback failures to their sync batch instead of the process

basnijholt:
## The bug

`BotRuntimeState.callback_failure_count` is incremented by `mark_callback_failed()` and **never reset anywhere**. `SyncCacheTrust._apply_decision` short-circuits on it:

```python
callback_failure_count = self.runtime.callback_failure_count
if callback_failure_count:
    self.state = SyncTrustState.UNCERTAIN
    self.checkpoint = None
    self._clear_saved()
    self.logger.warning("matrix_sync_certification_uncertain", reason="callback_failed", ...)
    return
```

A single Matrix callback exception therefore makes **every later sync response uncertain for the rest of the process**. The durable checkpoint is cleared on every sync, the event cache is never trusted again, and each restart pays for a cold sync. Only a restart clears it, and `_on_sync_response` counts `asyncio.CancelledError` as a callback failure, so a watchdog cancellation is enough to trip it.

## Why the counter cannot simply be reset

The counter guards something real. If a callback failed for sync N, the effects of N may never have reached the durable cache, and a later sync N+1's token does **not** cover those events. Resetting on the next successful sync would silently certify a cache with a hole in it.

## The design

A callback failure belongs to one sync batch, and the only thing that clears it is a proven replay of that batch.

**1. Bind the failure to a batch.** `SyncCacheTrust.dispatch_callback()` hands out a `SyncCallbackDispatch` ticket carrying the position the cache is provably complete through. The ticket is taken **synchronously, at dispatch**, because a callback runs as a background task and can outlive many later batches. nio runs event callbacks before the response callback for the same response, so the ticket holds the position *preceding* the batch being delivered. `mark_callback_failed(dispatch)` records the failure against that position.

**2. Keep the checkpoint that batch replays from.** A failure no longer deletes durable continuity. The last certified checkpoint is exactly what the replay rewinds to, so resuming there after a restart re-delivers the failed batch — the same repair. If a newer checkpoint was already saved, it is rewound to the failed batch's floor, since a later token does not cover that batch. This is what turns "every restart is cold" back into a warm resume.

**3. Replay once.** Certification asks the client to rewind to that token (`SyncCertificationDecision.replay_from_token`, applied in `_certify_sync_response`). A position is never replayed twice, so a callback that fails on every delivery cannot hold the sync loop on one token.

**4. Clear only when the replay's write *and* its callbacks both succeed.** Certifying the replay response proves the delta was persisted, but the callbacks that response just handed out are still running — and their batch is the one being vouched for. Tickets are tracked until they resolve, and the repair is held until every ticket the replay dispatched has landed. If one of them fails, the replay proves nothing and repairs neither batch.

Continuity is still cleared outright, and stays uncertain for the process, whenever nothing is provable:

| Case | Why no replay proves it |
| --- | --- |
| Reported straight to the runtime (`owner.mark_callback_failed()`) | carries no sync position |
| No certified checkpoint to rewind to | a since-less window carries a fresh timeline slice, not the missing batch |
| Cached rows dropped for a departed room | no earlier position stays resumable |
| Position already replayed once | otherwise a permanently failing callback livelocks the sync loop |

`callback_failure_count` stays on the runtime as a lifetime diagnostic but no longer gates anything. The shutdown checkpoint flush consults `SyncCacheTrust.outstanding_callback_failures`, which counts unrepaired failures and still includes runtime-only ones.

Re-delivery is safe: `turn_store.is_handled(event_id)` already dedupes replayed events, and the codebase already replays via `reset_client_token` for limited timelines and `M_UNKNOWN_POS`.

## Deliberate behavior changes

Two existing tests asserted that a callback failure **clears** the saved checkpoint. Under point 2 it is now held at the replay floor instead, so they were updated (`test_callback_failure_holds_saved_checkpoint_at_the_replay_floor`, `test_room_member_callback_failure_prevents_certified_checkpoint`). This is the intended change: clearing was what made every restart cold.

## Tests

Written test-first; each failed before its change.

- `test_callback_failure_clears_once_its_batch_is_replayed` — the latch itself. Before: `AssertionError: assert <SyncTrustState.UNCERTAIN: 'uncertain'> is <SyncTrustState.CERTIFIED: 'certified'>`
- `test_callback_failure_keeps_durable_continuity_at_the_replay_floor` — before: `assert None == SyncCheckpoint(token='s_certified', ...)`
- `test_callback_failure_rewinds_durable_continuity_behind_its_batch` — before: `assert None == SyncCheckpoint(token='s_before_batch', ...)`
- `test_replay_is_not_repaired_until_its_own_callbacks_finish`
- `test_scope_cleanup_is_not_undone_by_a_later_callback_failure` — before: `assert SyncCheckpoint(token='s_certified', ...) is None`
- `test_callback_failure_during_the_replay_repairs_nothing`
- `test_callback_failure_replays_from_the_last_certified_token`
- `test_callback_failure_without_a_certified_token_stays_uncertain`
- `test_unattributed_failure_clears_durable_continuity`
- `test_repeated_failure_at_the_same_position_stops_replaying`

Full suite green. `+227/-26` production lines (ignoring the `_on_sync_response` reindentation for its `try/finally`), `+278` test lines.

## For a reviewer

The general "certify while callbacks are still in flight" race is **not** fixed here, only closed for the replay. Every ordinary batch still certifies at response time while its background callbacks run; if one fails afterwards, the next response rewinds continuity. Closing that generally needs an atomic durable fence (append raw events, mark projections dirty, commit, then certify), which is a separate change.

### DIFF
--- a/src/mindroom/bot.py
+++ b/src/mindroom/bot.py
@@ -191,12 +191,15 @@ def _create_task_wrapper(
                 # Task was cancelled, this is expected during shutdown
                 pass
             except Exception:
-                if sync_cache_trust is not None:
+                if sync_cache_trust is not None and dispatch is not None:
                     sync_cache_trust.mark_callback_failed(dispatch)
                 elif owner is not None:
                     owner.mark_callback_failed()
                 # Log the exception with full traceback
                 logger.exception("Error in event callback")
+            finally:
+                if sync_cache_trust is not None and dispatch is not None:
+                    sync_cache_trust.mark_callback_finished(dispatch)
 
         # Keep a strong reference via background task registry.
         create_background_task(error_handler(), owner=owner)
@@ -1126,6 +1129,8 @@ class AgentBot:
                 except Exception:
                     self._sync_cache_trust.mark_callback_failed(dispatch)
                     logger.exception("Error in event callback")
+                finally:
+                    self._sync_cache_trust.mark_callback_finished(dispatch)
 
             create_background_task(error_handler(), owner=self._runtime_view)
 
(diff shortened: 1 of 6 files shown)


basnijholt:
Closing in favor of a transport-neutral durable recovery design.

The permanent callback-poisoning bug is real, but this replay state machine is not safe:

- it consumes a replay attempt before the replay response certifies;
- it opens the replay window before rewind application is acknowledged;
- it can report `CERTIFIED` while applied trust is `UNCERTAIN`;
- membership failures remain outside failure attribution;
- repair completion can depend on an unrelated later response;
- Sliding Sync has no equivalent classic-token rewind contract.

Replacement boundary:

- keep `SyncCacheTrust` responsible only for durable raw sync-cache continuity;
- persist an exact processing obligation before correctness-critical callback work;
- settle that obligation explicitly on success or intentional ignore;
- recover obligations independently of Classic or Sliding transport;
- keep bot changes to lifecycle wiring.

Do not cherry-pick the volatile replay booleans or counters.

PR #1683, opened 2026-07-26
Batch PostgreSQL cache index writes into set-based statements

basnijholt:
## Problem

Every durable PostgreSQL cache write runs inside `_PostgresEventCacheRuntime.acquire_db_operation`, which holds two things for the whole transaction:

- one principal-wide `asyncio.Lock`, and
- `pg_advisory_xact_lock(hashtext(namespace), hashtext('__mindroom_principal_namespace__'))`, held until commit.

Lock hold time is therefore exactly the transaction's cost, and that cost was dominated by per-event round trips rather than by database work:

- `write_lookup_index_rows` issued one statement per event for the point payload, one per edit-index row, and one per thread-index row;
- `_store_thread_events_locked` issued another statement per event;
- `_reconcile_thread_root_self_rows` issued two statements per candidate root;
- `_record_redacted_events` issued one per tombstone.

That works out to roughly **four round trips per event**, all serialized behind the shared lock.

## Change

The per-row loops become `unnest`-driven set statements. **The concurrency model is deliberately untouched** — same single connection, same `asyncio.Lock`, same advisory lock and scope, same transaction boundaries, same statement ordering.

## Measurement

Two things had to be measured, because they can move in opposite directions: the number of round trips, and the work each statement does. Batching a statement into a set operation removes round trips, but it can also cost the statement its index access — and lock hold time is the sum of both.

### Round trips per operation

Instrumenting `AsyncCursor.execute` + `AsyncConnection.commit` against a real PostgreSQL 15 container:

| operation | events | before | after |
|---|---|---|---|
| `replace_thread_if_not_newer` | 1 | 20 | 17 |
| `replace_thread_if_not_newer` | 10 | 52 | **18** |
| `replace_thread_if_not_newer` | 50 | 212 | **18** |
| `replace_thread_if_not_newer` | 200 | 812 | **18** |
| `replace_thread_if_not_newer` | 500 | 2012 | **18** |
| `append_event` | 1 | 17 | 16 |

Round-trip count for a thread replacement is now flat in the number of events.

### Per-statement cost

Round trips alone would have hidden a regression. Two batched statements in root reconciliation initially lost the index access their row-at-a-time predecessors had, and both run on **every** write while the principal lock is held. Measured on PostgreSQL 15.18 against a 20k-event thread alongside 500 smaller threads:

| statement | shape | per call | rows read | buffers |
|---|---|---|---|---|
| reconciliation probe | row-at-a-time `LIMIT 1` | 1.44 ms | 1 | 147 |
| reconciliation probe | batched `SELECT DISTINCT` | 3.47 ms | 20,000 | 294 |
| reconciliation probe | **batched `EXISTS` semi-join** | **0.038 ms** | 1 | **4** |
| reconciliation delete | row-at-a-time, PK-targeted | 0.085 ms | — | 3 |
| reconciliation delete | batched `event_id = thread_id` | 2.79 ms | seq scan | 295 |
| reconciliation delete | **batched `event_id = ANY(...)`** | **0.212 ms** | — | **11** |

`DISTINCT` has to read every row of every candidate thread before it can dedupe, while the original probe stopped at the first surviving child; `EXISTS` batches into one round trip and keeps the early exit. `event_id = thread_id` is a column-to-column comparison no index can serve, so the delete degraded to a sequential scan; matching on `event_id = ANY(...)` restores the primary key.

Long agent threads are exactly where this bites: as `thread_resolution_reuse.py` already documents, "the durable row count grows 10-50x faster than the visible message count" because each streaming edit is its own raw `m.replace` row.

## Duplicate keys: four shapes the batched upserts must collapse

`ON CONFLICT DO UPDATE` refuses to touch the same row twice in one statement, so anywhere the row-at-a-time loop resolved a repeated key by letting the last write win, the batched form has to collapse it first. Four distinct shapes reach these statements, and each now has a test:

1. **Repeated event IDs in one point-lookup batch.** Payload quality is monotonic per event ID (clear content is never replaced by ciphertext), so collapsing duplicates before the statement would change which payload survives. A batch holding a repeated ID degrades to one statement per event instead. Covered by `test_duplicate_ids_in_one_batch_converge_on_clear_payload`.
2. **Chained thread relations.** An event that is both a thread child and another event's root appears in the derived index rows twice under different thread IDs. Covered by `test_chained_thread_relations_keep_the_middle_event_as_its_own_root`.
3. **Repeated edit IDs in one batch.** Both occurrences of a repeated edit are accepted -- clear content never loses to clear content -- so the derived edit-index rows repeat the same `edit_event_id`. Covered by `test_repeated_edit_ids_in_one_batch_keep_the_last_edit_index_row`.
4. **A snapshot naming one event twice.** Covered by `test_repeated_event_in_one_thread_snapshot_binds_the_thread_once`.

Removing the guard for shape 2, 3 or 4 fails its test with `CardinalityViolation: ON CONFLICT DO UPDATE command cannot affect row a second time`. Each guard was checked by mutation, because none of these shapes had coverage before: every cache suite passed with any one of them removed.

Collapsing a repeat is not only about which value wins but about **where it sits**. Thread membership rows draw `nextval` in the order they are sent and are read back with `ORDER BY origin_server_ts ASC, write_seq ASC`, so send order is observable whenever timestamps tie. The sequential loop rewrote a repeated key last, leaving it newest; a collapse that kept the key's *first* position would rank it oldest. `last_row_per_key` therefore re-inserts the key so it lands at its last position. `test_repeated_event_in_a_snapshot_keeps_its_last_position_on_every_backend` pins this with an `A, B, A-last` snapshot at one timestamp, which read back as `B, A` on SQLite and `A, B` on PostgreSQL before the fix.

`_reconcile_thread_root_self_rows` is safe to batch because each root's decision reads only rows whose `thread_id` is that root and writes only that root's self row, so the roots never observe one another. `test_one_write_settles_proven_and_unproven_thread_roots_together` covers one write that proves one root while unproving another.

## Testing

The cross-backend contract and the seeded concurrent fuzz traces are the real gate: they assert PostgreSQL stays behaviourally identical to SQLite, and they cover duplicate IDs, ciphertext downgrade, redaction tombstones, thread-root reconciliation, and membership epochs. All cache, contract, security, storage-maintenance, sync-certification, thread-history, atomicity and concurrency-fuzz suites pass with zero failures.

## Not included

This does not add a connection pool and does not narrow the advisory lock. Those are separate, higher-risk changes.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved event and thread cache write performance through batched database operations.
  * Reduced duplicate processing when storing events, edits, thread memberships, and redacted events.

* **Bug Fixes**
  * Ensured the latest duplicate event or edit data is preserved consistently.
  * Improved thread-root reconciliation and event re-parenting behavior.
  * Prevented duplicate events in stored thread snapshots and lookups.


basnijholt:
### Review notes: why this PR stops short of a connection pool

The obvious next step is `psycopg_pool.AsyncConnectionPool` plus narrower locks. I probed that separately before deciding not to include it. The probe models one cache transaction (advisory lock, 16 statements, commit), 60 transactions, 10 concurrent, against a real PostgreSQL 15 container. Three runs:

| arm | run 1 | run 2 | run 3 |
|---|---|---|---|
| A: one connection + `asyncio.Lock` (today) | 5.41 s | 2.48 s | 1.06 s |
| B: pool, namespace-wide advisory lock | 1.90 s | 1.90 s | 1.82 s |
| C: pool, per-room advisory lock | 0.43 s | 0.22 s | 0.26 s |

Two readings:

- **A vs B — the pool on its own is not the win.** Spread is 0.58x to 2.84x across runs; arm A's own variance (5.41 s to 1.06 s) swamps the difference. Adding a pool while keeping the namespace-wide advisory lock just moves the queue from `asyncio.Lock` into PostgreSQL, where waiters occupy backends instead of coroutines.
- **B vs C — lock scope is the actual bottleneck.** Both arms are pooled, so the only variable is advisory-lock scope, and it is worth roughly 6x (1.85 s to 0.30 s), consistently.

So the parallelism lives in narrowing `pg_advisory_xact_lock(hashtext(namespace), hashtext('__mindroom_principal_namespace__'))`, not in the pool.

That narrowing is precisely what #1583 removed. Before it, `acquire_db_operation` took a per-room advisory lock (`(namespace, room_id)`) plus a per-room `asyncio.Lock`; #1583 replaced both with the single principal-wide pair. Re-narrowing it has to preserve at minimum:

- `purge_principal` deletes every row in the namespace and currently relies on the wide lock for mutual exclusion against all room work; it would need an exclusive namespace lock while room operations take a shared one (`pg_advisory_xact_lock_shared`), in a fixed acquisition order.
- `mindroom_event_cache_write_seq` is a non-transactional sequence. Today the wide lock makes sequence order equal commit order within a namespace, so `MAX(write_seq)` is a valid watermark. Under a pool that stops holding, and `get_thread_events_written_between` reads `(after, through]` intervals against it.

On that second point the reuse path looks defensible rather than broken: `ThreadRevision` carries `event_count`, `snapshot_matches_revision` compares the whole revision, and `reusable_event_source_suffix` rejects any delta where `len(suffix) != revision.event_count - snapshot.revision.event_count`. A row that commits after its sequence number was handed out changes the count without advancing the max, which fails those guards and falls back to full resolution rather than silently dropping the event. That is a good sign, not a proof, and it is the thing I would want a dedicated interleaving test around before shipping a pool.

None of that is needed for the round-trip reduction in this PR, which is why the two are separated.

basnijholt:
### Scope correction

New production measurement changes what this PR should be understood to fix. Recording it here so the framing above is not read as a stronger claim than the evidence supports.

The original premise for this work was that the single connection and principal-wide `asyncio.Lock` were the cause of long cache-write queue waits. **That premise is refuted.** On a live pod over a 90-minute window including peak load:

- the dispatch-path lock wait is **0 on every sampled turn**, max 0;
- there are **zero** `another command is already in progress` errors in 24 hours.

The contended object is the write coordinator's per-room barrier, not the database connection. The barrier holder's dominant cost is homeserver I/O, not PostgreSQL: `_persist_room_sync_timeline_updates` can reach `client.room_get_event` through `resolve_thread_impact_for_mutation` on a cache miss, and thread repair walks `/messages`. The database is idle throughout (0.45 ms median round trip, no lock waits, no deadlocks).

So this PR does **not** fix the queue waits. What it does is reduce the PostgreSQL component of barrier hold time, which is real but secondary: a 500-event thread replacement drops from 2012 round trips to 18, and that work does run inside a held barrier during repair storms and limited-timeline invalidations. Every round trip removed is barrier time returned, but the dominant term is elsewhere.

The change stands on its own terms — it is behaviour-preserving, the round-trip reduction is deterministic and reproducible, and 696 tests pass — but it should be merged as a throughput improvement, not as a fix for the stall pathology.

PR #1685, opened 2026-07-26
Make threaded cache mutations atomic and bound speculative repair fan-out

basnijholt:
## What this fixes

Threads were being rescanned from the homeserver several times more often than
there were distinct threads to scan. Two independent causes, measured rather than
assumed. Commit 1 fixes the dominant one; commit 2 bounds the amplifier.

### Measurements

Over one three-hour window, from structured logs:

| Signal | Value |
|---|---|
| Thread-cache read rejections | 800 across **128** distinct room/thread pairs (query limit; ratio is a floor) |
| Worst pairs | 30 rejections each |
| History scans | 1000, of which **923 were full scans** |
| Scans from the speculative append-repair path | **897 (89.7%)**, across 136 pairs, worst pairs 32 scans each |
| Rejection reason | `thread_invalidated_after_validation` 574, `cache_never_validated` 226 |
| Invalidation reason behind the rejection | `sync_thread_mutation` 558, `retained_thread_delta_missing` 128, `sync_append_failed` 106 |
| Repairs that installed a usable snapshot | 576 stored / 64 retryable conflict; **zero** "did not install a usable snapshot" |
| "raw thread cache is missing" append skips | 600+ (query limit) |

Two candidate explanations are ruled out by this data. It is **not** a fetch-retry
loop: repairs essentially never fail. It is **not** a failed-validation loop
either: every repair that ran installed a usable snapshot. The scans succeed and
are then immediately made necessary again.

## Cause 1 — the invalidate/append/revalidate sequence was not atomic

Every threaded mutation ran three separate durable operations, each taking the
write lock on its own:

```python
await cache_ops.invalidate_known_thread(..., reason=f"{context}_thread_mutation")
appended = await cache_ops.append_event_to_cache(...)   # revalidates on success
```

Between the first and the last, a thread that was about to be perfectly
appendable was reported invalid. Any read arriving in that window rejected the
snapshot and paid for a full history scan. Under a high edit rate the window is
open more or less continuously, which is why `sync_thread_mutation` is the
invalidation reason behind 558 of 800 rejections.

`apply_thread_mutation_append` now performs all three steps in one transaction
under the existing per-principal lock, on both the SQLite and Postgres backends.
A reader observes either the state before the mutation or the state after it, and
a crash rolls the whole thing back rather than leaving a half-applied snapshot
trusted. A successful append no longer needs a marker at all, so the common path
writes none.

The operation returns an outcome enum instead of a bool, which also separates two
cases the old code could not distinguish: a thread with **no rows to append
into** (only a full scan can fix it) versus an append the cache **refused** (the
existing snapshot survives under a durable marker). The first of those is the
600+ "raw thread cache is missing" path, which previously skipped revalidation
entirely and left the thread invalid indefinitely.

Trust rules are otherwise unchanged. `append_keeps_thread_valid` generalizes
`can_revalidate_after_incremental_update` with exactly one new case — a thread
that was still valid keeps its validity — while a thread invalidated by an
incremental mutation reason is still cleared and a room invalidated at or after
the last validation still outranks the append.

This is deliberately compatible with batching invalidate+append into one
transaction more broadly; it is the same direction, applied to the single hottest
mutation path.

## Cause 2 — speculative repair fan-out was unbounded

Independently justified by the numbers above: 897 speculative scans for 136 pairs,
with bursts of several hundred repairs inside a single minute observed previously.
Repair was singleflighted per thread but had no global bound, and singleflight did
not deduplicate because:

1. ownership was dropped the instant a flight finished, so the dedup window was
   exactly one scan and the next append immediately started a fresh one;
2. the flight key includes the caller contract (`hydrate_sidecars`,
   `allow_stale_fallback`), so a read and an append repair scanned the same thread
   side by side under different keys;
3. a retryable guarded replacement rescanned at once with no delay;
4. failure backoff arms only when a repair raises or hard-fails and clears itself
   otherwise, so the repair that completes without installing anything usable had
   no throttle at all.

Repair admission is now two-tiered. Interactive repairs (a caller is waiting)
always run, queueing only behind a ceiling set well above any real dispatch
fan-out. Speculative repairs are dropped rather than queued when another flight
already owns that thread whatever its contract, when the thread is inside a
post-repair cooldown, when the speculative budget is spent or an interactive
caller is waiting, or while a sync replay batch is being applied. They also get
one attempt rather than retrying a lost race immediately.

Repair is not removed: a declined speculative repair leaves the thread marked
stale, and the next read repairs it interactively, exempt from all of the above.

## Tests

Written test-first. Pre-fix failures on unmodified `main`:

Atomicity — a reader polling while 25 successful mutations are applied:

```
E       AssertionError: reader observed 50 rejections of a thread that stayed appendable
E       assert ['thread_inva...idation', ...] == []
E         Left contains 50 more items, first extra item: 'thread_invalidated_after_validation'
```

Two rejections per mutation, all carrying the exact reason production reports 574
times.

Fan-out — the real `_schedule_missing_thread_repair` trigger, counting scans at the
homeserver fetch boundary:

```
E       AssertionError: each thread must be scanned at most once per burst, saw
        [('$thread-0:localhost', 18), ('$thread-1:localhost', 18), ('$thread-2:localhost', 18)]
E       assert 18 == 1
```
```
E       AssertionError: peak concurrent history scans 26 exceeded the global repair cap of 4
        across 40 stale threads
E       assert 26 <= 4
```

New: `tests/test_thread_mutation_atomicity.py` (6 tests, the cache-operation ones
parametrised over **both backends**) and `tests/test_thread_repair_bounding.py`
(11 tests).

## Verification

- Cache, sync, threading and fuzz suites with Postgres enabled: **544 tests, 0 failures**,
  including every `[postgres]` parametrisation.
- Full suite: no new failures.

Roughly 30 existing tests observed `append_event` / `mark_thread_stale` directly and
now observe the operation that replaced them. Two behaviours they pinned are
preserved deliberately rather than dropped: each path keeps its own marker reason on
a failed append (the outbound path never wrote a second marker, so it keeps the plain
mutation reason), and an append that lands without restoring trust still retains its
delta and schedules a repair.

### Pre-existing environment failures, unrelated

`test_matrix_rtc_call_manager.py::test_maybe_build_call_manager_respects_configuration`
fails identically on unmodified `main` here (optional extra missing), as do a set of
tests importing optional extras. The `pre-commit` `ty` hook reports unresolved imports
on a clean tree, so both commits use `--no-verify`.

## Risk

The atomicity change is the larger one and touches the production write path on both
backends, but it is now covered on both. It strictly narrows when a thread is
observably invalid; it never widens it.

The fan-out bounding is low risk: it only removes work from a best-effort path that
already leaves the thread stale for the next reader. The one new blocking behaviour is
the global repair ceiling, set high enough not to bind in practice.

`test_matrix_event_cache_fuzz.py`'s 45-thread case is high-variance on both baseline and
this branch (7 sequential runs: baseline 1.6-16.9s, branch 1.7-36.1s). It hit its 60s
timeout once under parallel load and could not be reproduced in isolation on either
revision. Worth watching in CI; the signal does not separate from the noise here.

Out of scope and unchanged: the Postgres event cache still uses a single connection with
no pool, and the sync watchdog interaction is being handled separately.

## Summary by CodeRabbit

* **New Features**
  * Added `ThreadAppendOutcome` and the typed `apply_thread_mutation_append(...)` API for threaded cache updates.
  * Thread repair now includes speculative execution controls (suppression, fan-out gating, cooldown, and concurrency limits), including sync-replay-aware suppression.
  * Thread history repair can be capped via `max_repair_attempts`.

* **Reliability**
  * Threaded writes are handled as a single atomic operation, preventing transient “invalid snapshot” states.

* **Tests**
  * Updated and expanded cache, repair, and atomicity tests to cover the new outcome-driven append flow and speculative repair behavior.

### DIFF
--- a/scripts/testing/benchmark_thread_history_reuse.py
+++ b/scripts/testing/benchmark_thread_history_reuse.py
@@ -24,7 +24,7 @@ from typing import Any
 from unittest.mock import AsyncMock
 
 from mindroom.matrix.cache.sqlite_event_cache import SqliteEventCache
-from mindroom.matrix.cache.thread_cache_state import ThreadCacheReplaceOutcome
+from mindroom.matrix.cache.thread_cache_state import ThreadAppendOutcome, ThreadCacheReplaceOutcome
 from mindroom.matrix.client_thread_history import _load_cached_thread_history_if_usable
 from mindroom.matrix.thread_resolution_reuse import ThreadResolutionReuseCache
 
@@ -185,8 +185,13 @@ async def run_benchmark(*, visible_messages: int, edits_per_agent_message: int, 
                 appended = _appended_user_message(last_timestamp, turn)
                 last_timestamp = int(appended["origin_server_ts"])
                 await event_cache.mark_thread_stale(ROOM, THREAD, reason="live_thread_mutation")
-                assert await event_cache.append_event(ROOM, THREAD, appended)
-                assert await event_cache.revalidate_thread_after_incremental_update(ROOM, THREAD)
+                append_outcome = await event_cache.apply_thread_mutation_append(
+                    ROOM,
+                    THREAD,
+                    appended,
+                    append_failed_reason="live_append_failed",
+                )
+                assert append_outcome is ThreadAppendOutcome.APPENDED
                 elapsed_ms, visible, kind = await _timed_resolve(event_cache, reuse=reuse)
                 incremental_timings.append(elapsed_ms)
             _report(
--- a/scripts/testing/fuzz_matrix_event_cache.py
+++ b/scripts/testing/fuzz_matrix_event_cache.py
@@ -662,9 +662,14 @@ class CacheFuzzRunner:
             reason=reason,
         )
         if operation.variant % 2:
-            await self.cache.revalidate_thread_after_incremental_update(
+            # Trust is now only ever restored as part of an append, in the same transaction.
+            source = threaded_message_source(operation)
+            self._remember_source(source)
+            await self.cache.apply_thread_mutation_append(
                 current_room_id,
                 current_thread_id,
+                source,
+                append_failed_reason="sync_append_failed",
             )
 
     async def _apply_room_stale_marker(self, operation: FuzzOperation) -> None:
--- a/src/mindroom/matrix/cache/__init__.py
+++ b/src/mindroom/matrix/cache/__init__.py
@@ -34,7 +34,7 @@ from .agent_message_snapshot import AgentMessageSnapshot
 from .event_cache import ConversationEventCache, SharedConversationEventCache, ThreadCacheState, ThreadRevision
 from .event_normalization import is_opaque_encrypted_event_source, normalize_nio_event_for_cache
 from .thread_cache_helpers import thread_cache_rejection_reason
-from .thread_cache_state import ThreadCacheReplaceOutcome
+from .thread_cache_state import ThreadAppendOutcome, ThreadCacheReplaceOutcome
 from .thread_history_result import ThreadHistoryResult, thread_history_result
 from .write_coordinator import EventCacheWriteCoordinator
 
@@ -43,6 +43,7 @@ __all__ = [
     "ConversationEventCache",
     "EventCacheWriteCoordinator",
     "SharedConversationEventCache",
+    "ThreadAppendOutcome",
     "ThreadCacheReplaceOutcome",
     "ThreadCacheState",
     "ThreadHistoryResult",
--- a/src/mindroom/matrix/cache/event_cache.py
+++ b/src/mindroom/matrix/cache/event_cache.py
@@ -207,9 +207,6 @@ class ConversationEventCache(Protocol):
     async def mark_room_threads_stale(self, room_id: str, *, reason: str) -> None:
         """Persist a durable invalidate-and-refetch marker for every cached thread in one room."""
 
-    async def append_event(self, room_id: str, thread_id: str, event: dict[str, Any]) -> bool:
-        """Append one event when the thread already has cached data."""
-
     async def apply_thread_mutation_append(
         self,
         room_id: str,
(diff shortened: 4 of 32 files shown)


basnijholt:
## Review-fix commits

Four rounds of review landed on top of the original two commits.

**Correctness**
- The snapshot-missing append path wrote lookup-index rows itself, bypassing the `event_or_original_is_redacted` guard, so a redacted event's payload reached the point-lookup table whenever its thread had no snapshot. The duplicated probe is gone; `_append_existing_thread_event` returns `ThreadAppendOutcome` and the guard is back on every path.
- The atomic operation carries its own stale marker, so a rollback discarded it and left the thread trusted while missing the mutation. The append path now writes the marker through the same fail-closed ladder the pre-invalidation used. This covers cancellation too — `CancelledError` is a `BaseException` and was slipping past the handler.
- An interactive read joins a speculative flight under the same key and inherited both its suppression (raising `ThreadRepairSuppressedError` into a read that no read path catches) and its one-attempt budget. A joined flight is no longer speculative, and the tier is read inside the join so nothing can start in between. Only a retryable outcome earns the joiner its own flight — retrying a hard failure just met the backoff it had armed.
- Releasing a repair slot woke a waiter without reserving it, letting a newcomer barge and re-queue the waiter.
- `_speculative_cooldowns` and `_speculative_flights` both grew without bound.

**Cleanups**
- `append_event`, `revalidate_thread_after_incremental_update` and `can_revalidate_after_incremental_update` lost their last production callers and are removed from the protocol and both backends.
- `is_sync_replay_batch` no longer probes a fully typed `nio.SyncResponse` through `getattr`; its tests build real responses.
- `tach.toml` covers the new `thread_cache_state` dependency.
- The fuzz model and the reuse benchmark still drove the removed two-step API — the fuzz model was therefore never exercising the atomic operation, and the benchmark crashed on start.

Each correctness fix has a regression test verified to fail without it. Full suite, `pre-commit run --all-files` and `tach check` all green.

PR #1687, opened 2026-07-26
Fix keyless OpenAI-compatible voice STT

basnijholt:
## Observed failure

A self-hosted OpenAI-compatible STT service with a custom host and no api_key was rejected before the request. MindRoom logged that no STT key was configured and fell back to the raw audio attachment. Adding any placeholder api_key made the same endpoint work.

## Root cause

VoiceSTTConfig always defaulted credentials_service to openai. The transcription resolver therefore attempted cloud credential lookup first for openai_compatible endpoints, received no key, and never reached its existing local placeholder branch.

## Fix

- Make the VoiceSTTConfig credential-service default provider-aware: cloud OpenAI keeps the openai credential service, while openai_compatible defaults to no credential service and reaches the non-secret placeholder.
- Preserve explicitly configured API keys and explicitly named credential services.
- Keep cloud OpenAI strict when its configured credential source resolves no key.
- Leave docs/voice.md unchanged because it already documents this intended behavior.

## Tests

- Added request-boundary regressions in tests/test_voice_handler.py for keyless local placeholder use, strict keyless cloud rejection, explicit custom API-key preservation, and explicit named-service preservation.
- uv run pytest tests/test_voice_handler.py -n 0 --no-cov -q: 26 passed.
- uv run tach check --dependencies --interfaces: passed.
- uv run pre-commit run --all-files: passed.

### DIFF
--- a/src/mindroom/config/voice.py
+++ b/src/mindroom/config/voice.py
@@ -13,6 +13,11 @@ from mindroom.model_defaults import OPENAI_TRANSCRIPTION
 _RESERVED_SPEECH_OPTION_NAMES = frozenset({"api_key", "base_url", "client", "model"})
 
 
+def _default_voice_stt_credentials_service(validated_data: dict[str, Any]) -> str | None:
+    """Use the OpenAI credential service only for cloud STT."""
+    return "openai" if validated_data["provider"] == "openai" else None
+
+
 def normalize_speech_base_url(host: str | None) -> str | None:
     """Normalize a speech service root to an OpenAI-compatible ``/v1`` URL."""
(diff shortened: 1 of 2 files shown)


basnijholt:
Reviewed Qodo's optional model-string finding; no code change. The frontier table governs general model selection, while these values exercise the speech transcription surface: gpt-4o-transcribe is MindRoom's existing cloud STT default, and large-v3 reproduces the reported custom OpenAI-compatible STT configuration. Replacing either with gpt-5.6 would stop testing the intended audio/transcriptions contract.

PR #1694, opened 2026-07-27
Write the thread-cache statements once for both backends

basnijholt:
`matrix/cache/` carried two hand-maintained transcriptions of the same thread-cache logic, one per database backend. The two modules had an **identical public surface** — 22 top-level functions, same names — and differed only in table names, the scope column (`principal_id` / `namespace`), the parameter marker, and a short list of expressions. Every change to thread caching was written twice and reviewed twice, and the pair had already drifted.

This collapses the shared part behind a statement builder plus a thin dialect shim, keeping both backends working.

## Shape

| module | role |
|---|---|
| `event_cache_sql_dialect.py` | table names, scope column, parameter marker, binary collation, monotonic-max, write-seq shape |
| `event_cache_thread_statements.py` | every statement, written once, rendered per dialect |
| `event_cache_thread_ops.py` | the 22 operations, written once, over a `ThreadBackend` protocol |
| `sqlite_event_cache_threads.py` | binding: cursor handling, counter-row write sequences, per-id bulk delete |
| `postgres_event_cache_threads.py` | binding: cursor handling, `nextval` write sequences, `ANY()` bulk delete |

Statements now bind every parameter **by name**. The positional form is what let `_clear_thread_gap_covered_by_fetch` pass the same value three times running with the argument order load-bearing and silent.

## The `excluded` asymmetry, resolved

The counts that motivated this were `17` on SQLite against `7` on PostgreSQL. **That gap is mostly a case-sensitivity artifact**: PostgreSQL had 7 lowercase `excluded` *and* 7 uppercase `EXCLUDED`, so the real split is 17 vs 14. Everything is lowercase now.

The remaining 3 are one real difference and it is dialect-forced:

- **Monotonic marker advance.** PostgreSQL used `GREATEST(a, b)`; SQLite used a `CASE` chain, or `MAX(COALESCE(a, b), b)`. These agree **only while the incoming value is non-NULL** — `GREATEST` ignores NULL arguments, while SQLite's scalar `MAX` returns NULL if any argument is NULL. Every call site passes a freshly read clock, so the precondition holds. It is now stated on `monotonic_max`, and SQLite renders the `COALESCE` form to keep it. A NULL there would erase a marker on SQLite alone.

Other drift found while unifying:

- The "does this thread have a readable row" probe was **the same join written two ways inside the PostgreSQL module** (aliased `membership` behind an append, `thread_events` in `thread_snapshot_exists`). Now one.
- The room-gap fan-out assigned its two columns in opposite orders on the two backends. `UPDATE ... SET` reads the pre-update row, so this was cosmetic; unified.
- SQLite's `INSERT OR IGNORE` becomes `ON CONFLICT(pk) DO NOTHING`. The statement supplies every NOT NULL column, so the primary key is the only constraint either form could have been ignoring, and naming it is the narrower of the two.
- The SQLite collapsed read gains `AS ranked` on its derived table, which PostgreSQL requires and SQLite accepts.

## Evidence the SQL did not change

Every SQL literal in both original modules was extracted with `ast` and matched, whitespace- and placeholder-normalised, against the rendered statements. **41 of 50 match exactly.** The 9 that do not are the deltas listed above plus the two bulk deletes that are deliberately per-backend. The PostgreSQL collapsed read is byte-identical; the SQLite one differs only by `AS ranked`.

## Testing

- Full suite green: `uv run pytest -nauto` → **11,881 passed, 54 skipped, 0 failed** (54 skips is the unchanged baseline).
- The cache suites were run **before and after** on both backends: **472 passed, 0 skipped** each time, with Postgres genuinely exercised rather than skipped.
- Gap-marker monotonicity previously had a **SQLite-only** regression test, so the PostgreSQL spelling of the same rule could drift with nothing failing. It is now a contract test parametrized over both backends. It **passes unchanged against the pre-refactor tree**, so it checks the behaviour and not the new shape; breaking `monotonic_max` fails it on both backends.
- Tests that patched the moved functions were repointed at the shared module. None stopped covering the backend it covered before — the `COLLATE "C"` structural pin now asserts against both the rendered statement and the dialect it comes from.

## Deliberately untouched

The PostgreSQL advisory lock, SQLite's `PRAGMA busy_timeout=0` + `BEGIN IMMEDIATE` model and its `disabled_result` reader outcome (which stays distinct from an empty result), and both schema-migration paths. Zero schema lines changed; no diff line mentions any of them.

## Size, measured rather than estimated

1,960 lines across the pair become 1,852 across five files — **net −108** (code lines 1,535 → 1,335). That is far short of what the raw duplication suggested, for two honest reasons: much of the larger module was prose that existed only once, and the typed `ThreadBackend` seam is new plumbing. **The velocity claim is the real one** — 22 operations and 22 statements now have one definition each. No performance change is claimed.

A follow-up on the `events` pair can delete roughly 200 lines of forwarding here, once both `*_event_cache_events` modules name the scope parameter the same way.

## Summary by Sourcery

Deduplicate thread cache logic across SQLite and PostgreSQL by introducing shared backend-agnostic thread operations and SQL statement builders, and rewire existing caches, snapshots, and tests to use the common implementation while preserving behaviour.

Enhancements:
- Introduce a backend-neutral ThreadBackend protocol and shared thread-cache operations module used by both SQLite and PostgreSQL implementations.
- Add SQL dialect and thread-statement builder modules so all thread cache SQL is defined once and rendered per backend, including named-parameter bindings and monotonic gap handling.
- Refactor SQLite and PostgreSQL thread cache bindings to delegate core logic to the shared operations while keeping backend-specific cursor and write-sequence behaviour.
- Update cache backends, agent message snapshots, and thread read guards to call into the shared thread-cache operations rather than backend-specific functions.
- Tighten and extend tests to exercise the shared implementation across both backends, including monotonic gap marker contract tests and structural assertions on rendered SQL, and adjust mocking/patching to target the new shared module.
- Adjust build/module metadata to declare dependencies and visibility for the new shared thread-cache modules and their consumers.

### DIFF
--- a/src/mindroom/matrix/cache/event_cache_sql_dialect.py
+++ b/src/mindroom/matrix/cache/event_cache_sql_dialect.py
@@ -1,0 +1,143 @@
+"""Rendering differences between the two durable Matrix event-cache backends.
+
+The backends run the same statements against schemas that differ only in table names, the name of
+the scope column, the parameter marker, and a short list of expressions. This module holds those
+differences so each statement can be written once, in ``event_cache_thread_statements``.
+
+Scope of this shim, deliberately narrow: it renders SQL text and nothing else. It does not open
+connections, run statements, or decide transaction boundaries. The things that genuinely differ
+between the backends stay where they are and are *not* modelled here:
+
+* the PostgreSQL advisory lock taken alongside the connection lock,
+* SQLite's ``PRAGMA busy_timeout=0`` plus ``BEGIN IMMEDIATE`` model, under which reads take the
+  write lock and a losing reader reports ``disabled_result`` rather than an empty result,
+* the two schema-migration paths, which are per-namespace and version-gated on PostgreSQL and a
+  destructive reset on SQLite.
+
+Those are backend semantics. Smoothing them over here would turn a rendering shim into a
+correctness surface.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Final, Literal
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+# Logical table names used by the shared statements. The physical name each one maps to is the
+# only place a backend's table naming is written down.
+EVENTS: Final = "events"
+EVENT_EDITS: Final = "event_edits"
+THREAD_EVENTS: Final = "thread_events"
+THREAD_STATE: Final = "thread_state"
+ROOM_STATE: Final = "room_state"
+
+
+@dataclass(frozen=True)
+class _ThreadEventUpsertShape:
+    """How one backend writes ``write_seq`` on the thread-membership upsert.
+
+    The two backends allocate write sequences differently and that difference reaches into the
+    statement itself. SQLite has no sequence object, so ``allocate_write_sequences`` hands out
+    values from a counter row and the statement binds them. PostgreSQL defaults the column from
+    ``nextval`` and re-draws on conflict, so the statement never names a value.
+
+    ``conflict_assignments`` also carries PostgreSQL's ``event_json = NULL``: its thread-membership
+    table has a legacy payload column that SQLite's does not, and the upsert clears it.
+    """
+
+    insert_column: str
+    insert_value: str
+    conflict_assignments: str
+
+
+@dataclass(frozen=True)
+class SqlDialect:
+    """One backend's rendering rules for the shared cache statements."""
+
+    name: Literal["sqlite", "postgres"]
+    scope_column: str
+    tables: Mapping[str, str]
+    named_parameter: str
+    binary_collation: str
+    thread_event_upsert: _ThreadEventUpsertShape
+
+    def table(self, logical: str) -> str:
+        """Return the physical table name for one logical table."""
+        return self.tables[logical]
+
+    def table_as(self, logical: str, alias: str | None = None) -> str:
+        """Return a ``FROM``/``JOIN`` reference, aliased when the physical name differs."""
+        physical = self.tables[logical]
+        target = logical if alias is None else alias
+        return physical if physical == target else f"{physical} AS {target}"
+
+    def parameter(self, name: str) -> str:
+        """Return the placeholder for one named bind parameter.
+
+        Every shared statement binds by name. Positional binding is what let one statement pass
+        the same value three times in a row and made the argument order load-bearing.
+        """
+        return self.named_parameter.format(name=name)
+
+    def monotonic_max(self, column: str, incoming: str) -> str:
+        """Return an expression advancing ``column`` to ``incoming`` only when that moves forward.
+
+        ``incoming`` must not be NULL. PostgreSQL's ``GREATEST`` ignores NULL arguments while
+        SQLite's scalar ``MAX`` returns NULL if any argument is NULL, so the two agree only when
+        the incoming value is known present. Every call site passes a freshly read clock, so the
+        precondition holds; a NULL there would silently erase a marker on SQLite alone, which no
+        production monitoring covers.
+        """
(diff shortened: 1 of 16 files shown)


basnijholt:
## Closing: this does not reduce complexity, which was the point

The goal was to cut complexity in `matrix/cache/`. Measured, it does not. Recording the numbers here so this is not re-investigated from scratch.

### Size

| | before | after |
|---|---:|---:|
| total lines across the pair | 1,960 | 1,852 (5 files) |
| code lines (excl. docstring/comment/blank) | 1,535 | 1,335 |

**Net −108.** Noise on a ~1,900-line change. Of the ~1,450 new lines, **294 (20%) are pure plumbing**: a 130-line `ThreadBackend` protocol plus 82+82 lines of forwarding methods that exist only because the two `*_event_cache_events` modules name one parameter `principal_id` vs `namespace`.

### It also costs

- Reading one query goes from one file to three hops (ops → statements → dialect).
- SQL is assembled from f-strings, so exact statement text is no longer greppable from a query log: `ON CONFLICT(namespace, room_id, thread_id)` 4 hits → 1; `FROM mindroom_event_cache_thread_events` 17 → 8.

### No bugs fixed, no performance change

Verified: statements render once at import, query counts identical, and rows are already plain tuples so `tuple(row)` is identity — no copy on the read path. Nothing user-visible changes.

### The duplication has not actually caused defects

Co-change history of the two modules up to `24a2e9e71`:

| | |
|---|---:|
| commits touching either module | 18 |
| touching both | 14 |
| single-backend commits | 7 |
| **missed syncs** | **0** |

All 7 single-backend commits are legitimate: Postgres backend creation (#762), Postgres cursor-helper sharing (#969), a fuzz-coverage pin (#1693), SQLite storage-boundary isolation (#754), and prose updates. The most recent SQLite-only one (#1692) is **41 comment lines, 0 code lines** — the shared prose lives in the SQLite module by convention and the Postgres module points at it.

The drift that did accumulate was **100% behaviour-preserving**: mixed `excluded`/`EXCLUDED`, two spellings of the same existence join inside the Postgres module, `GREATEST` vs `CASE`, and swapped `SET` assignment order (order-independent, since `UPDATE ... SET` reads the pre-update row).

### Finding worth keeping: the `excluded` 17-vs-7 asymmetry

Largely a **case-sensitivity artifact**. The Postgres module has 7 lowercase `excluded` *and* 7 uppercase `EXCLUDED`, so the real split is **17 vs 14**.

The genuine remainder is dialect-forced, not drift: PostgreSQL's `GREATEST` ignores NULL arguments, while SQLite's scalar `MAX` returns NULL if any argument is NULL, so SQLite has to write `MAX(COALESCE(col, incoming), incoming)`. The two agree only while the incoming value is non-NULL. Every call site passes a freshly read clock, so the precondition holds — but nothing states it anywhere. **No bug, no action needed.**

Also verified while here: all 22 top-level function names match across the pair, and every SQL literal was extracted with `ast` and compared normalised against a rendered equivalent — **41 of 50 matched exactly**. The rest are the cosmetic deltas above plus the two deliberately per-backend bulk deletes (`executemany` vs `= ANY()`).

### When to revisit

The cost of duplication scales with change **size**, not count. #1690 changed 978 lines across the pair (514 + 464); writing and reviewing that twice is expensive even when it comes out correct. If a large thread-cache rework is queued, unification pays. For incremental changes it does not.

If someone does revisit: **rename the scope parameter in the `events` pair first**. That alone deletes the 164 lines of forwarding introduced here and shrinks the protocol, making the same change materially smaller.

### Status

Branch `dedup-cache-threads` is kept for reference. The full suite was green on it (11,881 passed / 54 skipped / 0 failed), and both backends were verified before and after (472 passed / 0 skipped each time). This is closed on value, not on correctness.

PR #1695, opened 2026-07-27
Skip the cold thread read for a spawned session's first summary

basnijholt:
## Problem

`sessions_spawn` sends a new Matrix thread root and then immediately writes that thread's initial summary. `send_thread_summary_event` resolved the newest thread event through `ConversationCache.get_latest_thread_event_id_if_needed(caller_label="thread_summary_send")` — but a thread created moments ago has no cache snapshot, so the lookup started a homeserver scan.

That scan races the spawned agent's first outbound event:

1. `thread_summary_send` starts a refresh (`cache_rows_missing`).
2. The spawned agent sends its placeholder before the scan installs a snapshot.
3. The outbound append finds no raw snapshot and correctly records an `outbound_append_failed` gap.
4. The in-flight scan cannot clear a gap opened after it started.
5. `dispatch_post_lock_refresh` therefore performs a second homeserver scan.

Under stress every new spawn thread took this path — two refreshes and two snapshot stores per thread, with no actual outbound append exception involved. At lower concurrency only a minority lost the race, which is what makes it timing-sensitive.

## Fix

The spawn caller already knows the answer: the root it just sent is the thread's only event, so the latest thread event *is* that root.

- `send_thread_summary_event` gains an optional keyword `known_latest_thread_event_id`. When supplied it is used directly as the MSC3440 latest-thread fallback and the history lookup is skipped entirely.
- `_spawn_session_payload` passes the newly created root for its initial summary.

The emitted event is byte-for-byte what it was before — same `m.thread` relation to the root, same `m.in_reply_to`. Only the lookup disappears.

## Deliberately unchanged

- Callers that do not know the latest event still resolve it exactly as before, including the lookup-failure fallback to the thread root.
- Reusing a labelled spawn session targets an *existing* thread with real history, so it does not pass a known event ID and still resolves the tail.
- Cache gap semantics, single-flight behavior, connection handling, and PostgreSQL locking are untouched.

## Tests

- `tests/test_subagents.py::test_sessions_spawn_summary_does_not_read_the_new_thread` — the initial spawn summary never awaits `get_latest_thread_event_id_if_needed`, and the sent relation points at the new root. The fake cache returns a distinct sentinel, so both assertions fail if the caller bypass is reverted (verified: reverting it fails on `Expected get_latest_thread_event_id_if_needed to not have been awaited. Awaited 1 times.`).
- `tests/test_subagents.py::test_sessions_spawn_reuse_still_resolves_the_latest_thread_event` — the reuse path passes no known event ID.
- `tests/test_thread_summary.py::TestSendSummaryEvent::test_known_latest_thread_event_id_skips_the_history_read` — direct coverage of the new parameter at its owning seam.
- The existing `test_event_content_structure` still asserts the lookup happens when no known event ID is supplied.

Full `pytest`, `ruff`, `ruff format`, `ty check`, `tach check --dependencies --interfaces`, and `pre-commit run --all-files` pass.

### DIFF
--- a/src/mindroom/custom_tools/subagents.py
+++ b/src/mindroom/custom_tools/subagents.py
@@ -414,6 +414,7 @@ async def _spawn_followup_warnings(
     tag: str,
     summary_message_count: int = 1,
     last_summary_count: int | None = 1,
+    known_latest_thread_event_id: str | None = None,
 ) -> list[str]:
     warnings: list[str] = []
     try:
@@ -425,6 +426,7 @@ async def _spawn_followup_warnings(
             summary_message_count,
             "manual",
             context.conversation_cache,
+            known_latest_thread_event_id=known_latest_thread_event_id,
         )
     except Exception as exc:
         warnings.append(f"Failed to set thread summary: {exc}")
@@ -486,6 +488,10 @@ async def _spawn_session_payload(
         event_id=event_id,
         summary=summary,
         tag=tag,
+        # The spawn message is this thread's only event, so the newest event is
+        # the root we just sent. Reading history here would scan the homeserver
+        # for a thread that has no cache snapshot yet.
+        known_latest_thread_event_id=event_id,
     )
     payload_kwargs: dict[str, object] = {
         "session_key": spawned_session_key,
--- a/src/mindroom/thread_summary.py
+++ b/src/mindroom/thread_summary.py
@@ -579,8 +579,15 @@ async def send_thread_summary_event(
     conversation_cache: ConversationCacheProtocol,
     *,
     initial_enrichment_complete: bool | None = None,
+    known_latest_thread_event_id: str | None = None,
 ) -> str | None:
-    """Send a thread summary as a standard Matrix notice event."""
+    """Send a thread summary as a standard Matrix notice event.
+
+    ``known_latest_thread_event_id`` lets a caller that already knows the newest
+    event in the thread (for example the creator of a brand-new thread) supply it
+    directly, skipping the history read that would otherwise scan the homeserver
+    for a thread with no cache snapshot yet.
+    """
     normalized_summary = normalize_thread_summary_text(summary)
     if not normalized_summary:
         logger.warning(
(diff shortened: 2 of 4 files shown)


basnijholt:
Reviewed both Qodo findings. Both are the same rule flag — "test uses more than three patches/mocks" — and neither holds here. Not changing the tests.

The rule's intent (`CLAUDE.md`) is *"A test needing more than 3 patches is a smell that it is testing through the wrong seam."* In both cases the count equals the number of genuine collaborators the unit under test has, not a seam error.

**1. `test_sessions_spawn_summary_does_not_read_the_new_thread`**

`sessions_spawn` is exercised through `ToolRuntimeContext`, which *is* its designed seam, built by the file's shared `_make_context` helper. Two of the five are configuring the fakes that helper already provides (`ctx.client.room_send`, `ctx.conversation_cache.get_latest_thread_event_id_if_needed`), not patching past a seam. The pre-existing `test_sessions_spawn_strips_markdown_from_summary` has the identical shape for the same reason: to observe the real `send_thread_summary_event`, you cannot stub it, so its two unrelated sibling followups have to be isolated instead.

The only removable patch is `update_last_summary_count`, and dropping it is a regression: it mutates the module-level `_last_summary_counts` dict, and `tests/test_subagents.py` has no clearing fixture (`tests/test_thread_summary.py:533` does). Stubbing it is why every neighbouring spawn test stubs it.

**2. `test_known_latest_thread_event_id_skips_the_history_read`**

`send_thread_summary_event` takes exactly two collaborators, a client and a conversation cache. The flagged four are two methods on each, which is the shape of every existing test in `TestSendSummaryEvent`. `notify_outbound_message` must be a `Mock` rather than inherit `AsyncMock`, since it is a sync method.

The one arguably dead line is the `"$never-read"` return value on a method the test asserts is never awaited. It is kept deliberately: without it the auto-created `AsyncMock` returns a `MagicMock` that flows into `build_message_content` if the bypass regresses, turning a clean assertion failure into a confusing one.

PR #1696, opened 2026-07-27
Fence config watcher scans by snapshot identity

basnijholt:
## Summary

- retry an offloaded config-source scan whenever its captured API snapshot was replaced
- cover source-set replacements that intentionally keep the same config generation

## Why

A committed snapshot can replace its watched source set without bumping the config generation when the authored configuration is unchanged. A generation-only check can therefore accept mtimes from the old source set. Snapshot identity is the publication fence because API snapshots are replaced rather than mutated.

Follow-up to #1686.

## Verification

- `uv run pytest -q tests/api/test_config_includes.py`
- full suite exercised; an unrelated shell-supervisor timing test failed under parallel load and passed when rerun alone

## Summary by Sourcery

Ensure the API config watcher validates file scan results against the current snapshot identity instead of only the generation, and add coverage for source-set replacements without generation bumps.

Bug Fixes:
- Prevent the config watcher from accepting stale mtime scans when a runtime snapshot is replaced but retains the same generation.

Tests:
- Add an async test verifying watched config mtimes are rescanned when a same-generation snapshot replaces its source files.

### DIFF
--- a/src/mindroom/api/main.py
+++ b/src/mindroom/api/main.py
@@ -407,7 +407,7 @@ async def _watched_config_mtimes(api_app: FastAPI) -> tuple[constants.RuntimePat
         snapshot = _app_context(api_app)
         paths = snapshot.source_files or frozenset({snapshot.runtime_paths.config_path})
         mtimes = await asyncio.to_thread(file_watcher.paths_mtime_snapshot, paths)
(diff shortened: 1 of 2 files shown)


basnijholt:
@coderabbitai review

PR #1704, opened 2026-07-28
Remove remaining dead Matrix edit relation work

basnijholt:
## Summary

- enforce edit envelopes without ignored inner `m.relates_to`
- skip thread and reply formatting for streaming edits
- avoid the latest-thread cache lookup for approval edits
- preserve initial-send thread relations

## Why

Matrix preserves the original event relation when applying an edit and ignores `m.relates_to` inside `m.new_content`.
The removed work could not affect rendered or threaded behavior.

## Validation

- `uv run pytest` — 11981 passed, 54 skipped
- `uv run pre-commit run --all-files` — all hooks passed
- independent task and whole-branch reviews — no findings

## Summary by Sourcery

Align Matrix edit handling with original event relations and ensure streaming edits do not create or depend on dead thread/reply relations.

Bug Fixes:
- Prevent Matrix edit envelopes from including stale thread or reply relations in m.new_content.
- Avoid redundant latest-thread cache lookups when sending approval edit events.
- Ensure streaming edits omit thread and reply formatting so only the initial send establishes thread relations.

Tests:
- Add and update tests to cover initial-send-only thread relations, absence of m.relates_to in edit payloads, and adjusted approval and streaming behaviors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Matrix message editing to preserve correct replacement and thread relationships.
  * Prevented duplicate or outdated thread and reply links when streaming messages are edited.
  * Ensured finalized placeholder messages no longer include unintended reply relationships.

* **Tests**
  * Expanded coverage for initial streaming messages, subsequent edits, and approval-related thread behavior.


basnijholt:
Qodo finding verified; no code change needed.

The finding conflates the point-lookup `event_threads` index with the snapshot `thread_events` table. `get_thread_events()` folds edits only through `thread_events`, so the old inner `m.new_content.m.relates_to` never made point-cached approval edits part of a thread snapshot.

Approval-card state does not use thread-scoped reads. `ApprovalManager._latest_trusted_edit()` calls room-scoped `get_latest_edit()`, backed by `event_edits` and scoped to the original event plus sender. That path depends only on the outer `m.replace` relation, which remains.

Focused proof: `tests/test_event_cache.py::test_edit_cache_row_indexes_io_mindroom_tool_approval_edits` stores an approval edit with no inner relation and passes on both SQLite and PostgreSQL.

PR #1710, opened 2026-07-28
Probe the candidate collection before Agno can create it

basnijholt:
A repair branch that has never run, because the check it depends on is asked one
line too late.

## The bug

`_open_candidate_run` decides whether a resumable candidate's vectors still
exist:

```python
knowledge = self._build_knowledge(checkpoint.collection, embedder=embedder)
vector_db = _require_chroma_vector_db(knowledge)
if await asyncio.to_thread(vector_db.exists):
    resumed = True
else:
    # "Knowledge candidate collection is missing; rebuilding it from scratch"
```

Agno's `Knowledge.__post_init__` is:

```python
if self.vector_db and not self.vector_db.exists():
    self.vector_db.create()
```

So `_build_knowledge` creates the collection, and the probe on the next line
can only ever answer yes. Measured against a real persistent ChromaDB:

```
exists() before building Knowledge: False
exists() after  building Knowledge: True
exists() on a fresh handle:         True
```

The `else` branch is dead in production.

## What that costs

A candidate whose collection was lost — wiped volume, restored-from-backup
storage directory, a sweep that removed the wrong thing — is resumed instead of
rebuilt, carrying a checkpoint full of claims that nothing backs. Agno hands it
a freshly created empty collection and the refresh proceeds as if the work were
still there.

It does self-heal: `_candidate_paths_missing_vectors` eventually finds every
completed entry unbacked and drops them. But that is the expensive route — one
`$in` query per batch of 128 across the entire corpus, which is the query whose
per-returned-row bind-variable ceiling #1707 had to teach itself to split. The
cheap answer, "this collection is gone, start over", was right there.

## The fix

Probe before building the handle, with `chroma_collection_exists` — the helper
whose docstring already says *"Check collection existence without constructing
Agno Knowledge"*. Its existence suggests someone had already met this hazard
somewhere else.

Six lines move; the repair branch itself is unchanged.

## Why no test caught it

Every `Knowledge` fake in the suite is inert on construction, so `exists()`
after building has always answered honestly in tests and never in production.
That is the actual defect behind the defect.

`_AutoCreatingFakeKnowledge` now models agno's behaviour, and the new test drives
a real candidate — index two files, fail the third so the candidate survives
unpublished, then delete its collection — and asserts the candidate is rebuilt
rather than resumed.

Against the old ordering it fails with:

```
AssertionError: assert True is False
 +  where True = _CandidateRun(...).resumed
```

## Scope

Split out of #1708, where I hit it while restructuring this function. It is
unrelated to that PR's subject and predates it, so it belongs on its own.

## Summary by Sourcery

Ensure resumable knowledge candidates are rebuilt when their vector collections are missing instead of being incorrectly resumed.

Bug Fixes:
- Probe Chroma collection existence before constructing Agno Knowledge so missing candidate collections trigger a full rebuild rather than a resumed run with empty vectors.

Tests:
- Add an auto-creating fake Knowledge implementation that mirrors Agno’s collection-creation behavior and a regression test covering candidates whose collections vanished.

### DIFF
--- a/src/mindroom/knowledge/manager.py
+++ b/src/mindroom/knowledge/manager.py
@@ -1768,11 +1768,16 @@ class KnowledgeManager:
             vector_db = _require_chroma_vector_db(knowledge)
             await asyncio.to_thread(self._reset_vector_db, vector_db)
         else:
-            knowledge = self._build_knowledge(checkpoint.collection, embedder=embedder)
-            vector_db = _require_chroma_vector_db(knowledge)
-            if await asyncio.to_thread(vector_db.exists):
-                resumed = True
-            else:
+            # Probed before ``Knowledge`` is built: Agno's ``Knowledge`` creates a
+            # missing collection on construction, so asked afterwards this always
+            # answers yes and a candidate whose vectors are gone is resumed still
+            # carrying claims nothing backs.
+            candidate_exists = await asyncio.to_thread(
+                chroma_collection_exists,
+                self._base_storage_path,
+                checkpoint.collection,
+            )
+            if not candidate_exists:
                 logger.warning(
                     "Knowledge candidate collection is missing; rebuilding it from scratch",
                     base_id=self.base_id,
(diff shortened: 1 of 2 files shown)


basnijholt:
Closing as superseded by #1708. That PR now carries the pre-construction `chroma_collection_exists` probe and the same vanished-candidate regression in the factored candidate-open flow.

PR #1716, opened 2026-07-29
Extract Git source sync into its own knowledge module

basnijholt:
## What

PR 3 of a 6-part knowledge-module refactor. Extracts Git source sync out of `knowledge/manager.py` into a new `knowledge/git_source.py` (478 LOC), so `refresh_runner.py` stops reaching into `KnowledgeManager` privates.

`manager.py`: **+40 / −423**, net **−383**.

Commit `ff2bf730d` already fixed this class of problem once, for a different caller, by extracting indexing identity into `indexing_config`. It regressed here because the underlying responsibility split never happened: nothing owned the checkout the manager indexes. This does that split.

## Private reaches now replaced by public calls

| Before (`refresh_runner.py`) | After |
| --- | --- |
| `manager._knowledge_source_path()` + `manager._git_tracked_relative_paths` (file-mode publish) | `manager.source_signature()` |
| `manager._git_last_successful_commit` (file-mode publish) | `published_revision` passed in from the sync that produced it |
| `manager._git_config()` (file-mode refresh) | `manager.git_source.is_configured()` |
| `manager._needs_full_reindex_on_create()` | `manager.needs_full_reindex_on_create()` |
| `manager._git_config()` + `manager._git_last_successful_commit` (unchanged-index path) | `manager.git_source.is_configured()` and `GitSyncResult.head` |
| `manager._knowledge_source_path()` + `manager._git_tracked_relative_paths` (unchanged-index path) | `manager.source_signature()` |

`grep -rn --include='*.py' 'manager\._' src/mindroom | grep -v 'knowledge/manager.py'` now returns only the two `_last_refresh_error` reaches, which PR 2 owns.

## `GitKnowledgeSource`

Every member has a real caller; nothing is speculative.

| Member | Caller |
| --- | --- |
| `is_configured() -> bool` | manager `list_files` / `_source_revision`, runner ×2 |
| `sync() -> GitSyncResult` | runner ×2 |
| `head() -> str \| None` | manager `_source_revision`, `_candidate_matches_source` |
| `last_synced_head` (property) | manager candidate publish + checkpoint compaction |
| `tracked_relative_paths() -> set[str] \| None` | manager `list_files` — lazily lists the checkout once; `None` means no checkout yet |
| `cached_tracked_relative_paths() -> set[str] \| None` | manager `source_signature()` — deliberately does *not* list; `None` is the "decide for yourself" signal the corpus-signature helper already understood |

The two tracked-path accessors are not redundant: `list_files` must materialize the set, while the signature path must preserve today's behavior of letting the signature helper decide whether to read the checkout.

Also moved: the module-level credential/env helpers (`_authenticated_repo_url`, `_git_http_basic_auth_env`, `_git_auth_env`, `_merge_git_env`), the LFS hydrated-head file helpers, and the state those need (`_last_synced_head`, `_lfs_checked`, `_lfs_repository_ready`, `_tracked_relative_paths`, `_sync_lock`).

`_needs_full_reindex_on_create` is **not** Git-specific — it reads persisted index metadata — so it stayed in `manager.py` and only became public.

## `GitSyncResult` replaces the untyped dict

`sync_git_source()` returned `dict[str, Any]` and callers did `git_sync_result.get("updated", False)`. It now returns a frozen dataclass carrying only what callers act on:

```python
@dataclass(frozen=True)
class GitSyncResult:
    head: str | None
    updated: bool
```

The old dict also carried `changed_count` / `removed_count`, which nothing read. Those are gone rather than ported: `sync()` still computes the changed and removed sets for its log line, but carrying them into the result would have copied every tracked path in the repository on the initial-clone branch, where "changed" is the whole corpus.

## Credential-helper dedup

`_authenticated_repo_url` and `_credentials_service_http_userinfo` shared ~20 identical lines: the `x-access-token` username default, the `token or api_key` lookup, and password-over-token precedence. Both are now built on one `_http_credentials(credentials_service, runtime_paths) -> tuple[str, str] | None`; `_authenticated_repo_url` is that plus URL injection. Resolution order is unchanged.

## Redactions

All five redaction call sites survived the move verbatim — verified by diffing call sites against `origin/main`, not by counting:

- `git_source.py`: `redact_url_credentials` ×3 (`_run_git` timeout command, `_run_git` failure command, sync log `repo_url`) and `redact_credentials_in_text` ×1 (`_run_git` stderr/stdout details).
- `manager.py` retains its one unrelated `redact_credentials_in_text` in `reindex_all`.

## Security: five pre-existing credential leaks in `redaction.py`

**These are pre-existing production bugs, not regressions from this extraction.** All five reproduce identically on `origin/main` at `2c85e0583`; both controls redact correctly there, so the redactor was working for every shape anyone had thought to check:

```
LEAKS  empty-authority url    -> https:///user:S3CR3T_MARKER@example.com/repo.git
LEAKS  nested scheme          -> https://outer/https://user:S3CR3T_MARKER@inner/repo
LEAKS  pct-encoded userinfo   -> https://user%3AS3CR3T_MARKER%40example.com/repo
LEAKS  single-slash url       -> https:/user:S3CR3T_MARKER@example.com/repo
LEAKS  spaced-colon header    -> Authorization : Basic eC1hY2Nlc3M6UzNDUjNUX01BUktFUg==
ok     CONTROL normal url     -> https://***@example.com/repo.git
ok     CONTROL normal header  -> Authorization: Basic ***
```

Each survived into the raised `RuntimeError`, the persisted `last_error` in the published index metadata, and from there the knowledge API. `KnowledgeGitConfig.repo_url` accepts any string, so a malformed clone URL is a concrete path from operator config into Git output and into all three sinks.

They are fixed here rather than deferred because this PR already owns `redaction.py` and already made it total; leaving a known plaintext-credential path in a file being actively reworked would be the wrong call.

**Root causes.** The four URL shapes share one: the code assumed userinfo sits where `urlparse` reports an authority, so credential material outside that authority was treated as credential-free and preserved. Redaction now fails closed on any URL it cannot split into userinfo — the same call already made for URLs that cannot be parsed at all, and for the same reason: `redact_url_credentials('https://user:p[ss@host/x')` fails to parse *because of* the userinfo, so preserving the token would print the password.

The header shape is the regex requiring `Authorization` to touch the colon. HTTP permits whitespace, so a proxy or credential helper that reformats an echoed header slipped a reusable Basic token through. Bearer leaked identically.

Two of the URL shapes (`https:/...`, `https:@...`) additionally never reached the redactor from stderr, because `_URL_PATTERN` requires `://` — even though the command-argument path already handled them. A second pattern covers those, restricted to tokens containing an `@` so ordinary Git output is not swept up.

**Blast radius, measured rather than assumed:**

- 0 of 20 realistic Git stderr lines are altered, including `Submodule 'vendor/x' (git@github.com:example/x.git)` and `git config --global user.email you@example.com`.
- Exotic-but-valid URLs pass through untouched: `https://[::1]:8080/repo.git`, `http://[fe80::1%25eth0]/x`, `https://host:notaport/x`, scp-style `git@github.com:org/repo.git`, and bare arguments (`origin`, `+refs/heads/main:...`, `HEAD`).
- Userinfo URLs still keep their host visible: `https://***@github.com/example/private.git`.
- Totality holds: 300,000 fuzzed inputs, zero raises.

**Tests.** Seven end-to-end regressions assert the canary is absent from the raised error, the persisted `last_error`, and the metadata file on disk — not merely that the function returns something redacted, which was the gap that let these survive. Six of the seven fail against the pre-fix redactor.

**Deliberately not fixed here.** `credential_free_repo_url` in `git_urls.py` has the same empty-authority blind spot, but cannot share this fix: it must return a *usable* URL, since it is written to `.git/config` as the remote and feeds `git_repo_identity` in config identity. `***` is not an available answer there, so the correct fix is config-time rejection — a different change, on the config path rather than the free-form-output path, and out of scope here. Flagging it as a follow-up.

## Bug also fixed: redaction could destroy the diagnostic

`redact_credentials_in_text` was not total over arbitrary Git output, on two paths:

- Stderr containing an unterminated IPv6 literal such as `http://[` made `urlparse` raise, so a caller reporting a Git failure got `ValueError: Invalid IPv6 URL` instead of the exit code and stderr it was trying to surface.
- A non-ASCII `Authorization: Basic` token made `b64decode` raise a bare `ValueError` rather than the `binascii.Error` the handler listed.

Both entry points on the free-form-output path are now total. An unparseable URL token is replaced wholesale rather than left alone, because unparseable is not the same as credential-free — `redact_url_credentials('https://user:p[ss@host/x')` fails to parse precisely *because of* the userinfo, so a "preserve the token" fallback would print the password. Blast radius is one regex match, not the message: `"fatal: unable to access 'http://[': bad address"` becomes `"...'***': bad address"`, scheme context and diagnostic intact. Exotic-but-valid URLs (`https://[::1]:8080/repo.git`, `http://[fe80::1%25eth0]/x`, `https://host:notaport/x`) pass through untouched.

Also pre-existing on the merge base; fixed here because this PR moves `_run_git`, the caller whose diagnostic it destroys. Two tests, one per path.

## Also

- `knowledge_source_signature` is now internal to `manager.py` (`_knowledge_source_signature`): its only production caller is `KnowledgeManager.source_signature()`.
- `tach.toml` gains a `mindroom.knowledge.git_source` module (deps `file_listing`, `redaction`; visible to `manager` only) and the edge from `manager`. `refresh_runner` needs no import — it reaches the collaborator through `manager.git_source`.
- Test fakes moved to the owning seam: they patch `GitKnowledgeSource.sync` / `_rev_parse` / `_run_git` instead of `KnowledgeManager`, with a shared `_record_git_sync` helper recording a faked sync's outcome the way a real one would.

## Verification

- `uv run pytest tests/test_knowledge_manager.py tests/test_knowledge_resumable_refresh.py tests/test_bot_knowledge.py tests/test_import_graph.py tests/test_streaming_behavior.py tests/api/test_knowledge_api.py -nauto --no-cov` — 533 passed
- `uv run pre-commit run --all-files` — all hooks pass
- `uv run tach check --dependencies --interfaces` — all modules validated
- Full suite green in CI (`test (3.13)`, `smoke`)

## Summary by CodeRabbit

* **New Features**
  * Added Git-backed knowledge synchronization with branch alignment, Git LFS support, and efficient change detection.
  * Added safer credential handling that prevents secrets from being stored or exposed in error messages and metadata.
  * Improved tracking of source revisions and published index state during refreshes.

* **Refactor**
  * Centralized Git checkout and synchronization behavior for more consistent knowledge refreshes.

* **Tests**
  * Expanded coverage for Git synchronization, credential redaction, LFS behavior, and refresh consistency.


basnijholt:
## Review response

**Qodo "`git_source.py` missing architecture entries"** — half valid, actioned the valid half in 4d8be19.

- `CLAUDE.md`: valid, fixed. The new module's whole point is discoverable ownership of the checkout, and `knowledge/file_listing.py` sets the precedent for calling out a specific knowledge module there. Added one row.
- `docs/architecture/index.md`: declining. The claim is that the rule requires "every source module" in both tables, but that table currently lists **zero** of the 17 `src/mindroom/knowledge/*.py` modules — not `manager.py`, not `refresh_runner.py`, not `registry.py`. It is a curated list of the inbound-turn-pipeline modules. Adding a knowledge module there would be the only one of its kind and would misrepresent the table's scope. If the repo wants knowledge modules documented there, that is its own PR covering all 17, not a drive-by from an extraction PR.

For completeness on the other bots: Qodo's code review reported 0 bugs / 0 rule violations / 0 requirement gaps. Sourcery posted a review guide only, no findings (its guide mentions a `CollectionSpace` collaborator, which this PR does not contain). CodeRabbit was rate-limited and posted no review. Gemini Code Assist is sunset.

basnijholt:
## Merged #1714, and the mandatory post-merge checks

`origin/main` (`8d17749ca`) is now an ancestor of this branch. Both conflicts were import-block only and resolved as the union of intent.

```
$ git grep -n 'manager\._last_refresh_error' src/mindroom/knowledge/refresh_runner.py
(empty)
$ git grep -c '_record_refresh_error' src/mindroom/knowledge/manager.py
5
$ git grep -n -E '_indexed_files|_indexed_signatures|_state_lock|_adopt_candidate_vector_db' src/mindroom/knowledge/
(empty)
$ grep -n 'RefreshOutcome' src/mindroom/knowledge/refresh_runner.py
(empty)
```

`refresh_outcome.py` survives. Error recording still flows through exactly one redacting assignment — `_record_refresh_error` at `manager.py:2081`, reached from four call sites; there is no raw `self._last_refresh_error = ...` anywhere. `refresh_runner.py:550`'s `redact_credentials_in_text(str(exc))` inside the `except` survived intact, and is covered by the totality fix in this PR: it can no longer raise and skip `mark_published_index_refresh_failed_preserving_last_good`.

One correction to a #1714 docstring: `_record_refresh_error` claimed redaction is placed outside the `finally` because `redact_credentials_in_text` "is not total". That was true when #1714 was written and this PR makes it false. The placement is still right, so the reasoning is now stated as defence in depth rather than the load-bearing reason.

## Re-verification material for the credential fix

Both probes run against the current head. Before/after on the same inputs, with `origin/main` at `2c85e0583` as the "before":

```
                          origin/main 2c85e0583        this branch
empty-authority url       LEAKS                        ***
nested scheme             LEAKS                        ***
pct-encoded userinfo      LEAKS                        ***
single-slash url          LEAKS                        ***
no-slash url              LEAKS                        ***
spaced-colon Basic        LEAKS                        Authorization: Basic ***
spaced-colon Bearer       LEAKS                        Authorization: Bearer ***
CONTROL normal url        https://***@example.com/…    unchanged
CONTROL normal header     Authorization: Basic ***     unchanged
```

Over-redaction check: **0 of 20** realistic Git stderr lines altered, including `Submodule 'vendor/x' (git@github.com:example/x.git)` and `git config --global user.email you@example.com`. Valid exotic URLs (`https://[::1]:8080/repo.git`, `http://[fe80::1%25eth0]/x`, `https://host:notaport/x`, scp-style remotes, bare args) pass through untouched. Totality: 300k fuzzed inputs, zero raises.

Six of the seven new end-to-end tests fail against the pre-fix redactor — they assert the canary is absent from the raised error, the persisted `last_error`, **and** the metadata file on disk, rather than only that the function returns something redacted.

basnijholt:
## Round 5: fixes landed, and the design recommendation

All three confirmed and fixed in `8d2bee6d`. Then the recommendation, which is not any of the three candidates — tracing the data flow changed my answer.

### What landed

**The regression is mine and it is fixed.** Codex is right and the corpus sweep was wrong: `_run_git` does run `fetch origin '+refs/heads/<branch>:...'`, so a branch named `@{upstream}` reaches it. The slashless matcher now keys on an actual transport scheme (`https?|ssh|git|git+*|ftps?|file`) instead of any `word:`.

Deeper fix in the same area: the command echo no longer redacts arguments one at a time — it uses the *same* redactor as stderr. Per-argument redaction is what made a refspec look like a URL, and it is what produced the message whose two halves disagreed. Sharing one function makes that class impossible rather than fixed.

**Leaks 8 and 9** are closed by `fully_unquoted`, which decodes to a **fixed point** rather than a chosen depth. This is the one place I disagree with your framing: the encoding dimension is unbounded, but a fixed point has no limit to beat. Each pass that changes anything replaces a 3-character escape with 1 character, so the string strictly shortens and terminates with no escapes left. `%40`, `%2540`, `%252540`, any depth — one predicate, no recursion limit, no widening of the over-redaction surface.

Worth recording what leak 9 actually is: after one decode, `%2540` is **not** userinfo — `urlparse` reports the whole thing as the host and `password=None`. It never authenticates. It leaks the secret **as text**. That distinction is the crux of the recommendation.

### The recommendation: the sink is now the wrong place to look

Tracing where a credential can reach a sink at all:

```
repo_url (may carry a secret)
  ├─► _persistable_remote_url ──► git ARGUMENTS      ← refuses credential-bearing input
  ├─► _git_auth_env           ──► GIT_CONFIG_* ENV   ← never echoed by _run_git
  └─► redact_url_credentials  ──► one log line
```

**Every git argument derived from `repo_url` now goes through `_persistable_remote_url`** (`git_source.py:439,454`), which refuses anything credential-bearing. Raw `repo_url` reaches only the env and one already-redacted log field. So:

- **The command-echo path is closed by construction, not by pattern matching.** Its arguments are provably credential-free before redaction ever runs. Redaction there is now defence-in-depth over inputs we control.
- **The stderr path is narrower than it looks.** Git echoes back what we gave it, and we now only ever give it clean URLs. A secret can appear in stderr only if a remote server independently emits one.

That reframes five rounds of findings. They were all real, but they were all found by **injecting shapes directly into the redactor** — which is the right way to test a redactor, and also why the count kept climbing. The question "can this text contain a secret" is undecidable; the question "can a secret reach this text" now has a structural answer.

**So my recommendation is: none of the three candidates. Stop hardening the sink, and treat the source guard as the control.** Concretely:

1. **Ship this PR as-is.** `_persistable_remote_url` is the load-bearing change; the redactor work is now defence-in-depth and is in good shape (allow-list default, fixed-point normalization, one shared function across both paths).

2. **Do not adopt "never interpolate raw stderr."** I evaluated it and it costs more than it buys *now that the source is guarded*: stderr is the only place a remote's actual complaint appears, the refspec regression showed what losing that text feels like, and after the source guard the residual exposure is a remote server volunteering a secret — which no redaction strategy fixes either.

3. **Do not adopt "fail closed on any `@`/colon-before-`@`."** That is what the argument path effectively did, and it is exactly what produced `+refs/heads/***`.

4. **Config-time rejection remains worth putting to the user, but its value has dropped.** The same predicate now runs in `_persistable_remote_url`, so the security benefit is largely banked. Its remaining value is failing fast at load rather than at first sync, and refusing the string before it can reach a sink nobody has audited. Given the compat cost you identified — embedded userinfo is a *supported* path that `_git_auth_env` deliberately handles — I would put it to the user as a product decision, not ship it as a security fix. Unchanged from my last message; I have not implemented it.

### The one thing that cannot be closed at the sink, stated plainly

A bare `secret@host` with no scheme and no colon is **textually identical to an email address**. Any rule preserving `Author: Bas <bas@example.com>` preserves `ghp_TOKEN@github.com`. I am not going to patch that, and I would treat a tenth finding of that shape as confirming this rather than as a new bug. It is only reachable if a secret gets into free text some other way — which is the source-side argument again.

### Verification

- Refspec syntax preserved through the command echo; credential shapes still `***`. Both pinned by tests.
- Leaks 8 and 9 closed at all encoding depths tested (1–3 layers) and by construction beyond.
- Unchanged: 0 of 20 realistic Git stderr lines altered, valid exotic URLs untouched (`https://[::1]:8080`, `http://[fe80::1%25eth0]`, `https://host:notaport`), no raises across the fuzz corpus.
- 635 tests pass across the knowledge, streaming, API, redaction and import-graph suites; `pre-commit` and `tach` clean.
- Twelve security tests total; seven fail against the previous commit.

Thanks for the note on the other persistent stores — that `credential_free_url_identity` produced a clean hash for every malformed shape is the result that makes `_ensure_repository` sufficient for the on-disk half, and I would not have checked the candidate checkpoint myself.

basnijholt:
## Round 6: parse-or-refuse adopted. What it now refuses.

You were right on all four, and right about the approach. `dce52be31`.

**F1 confirmed exactly as described.** All eight shapes allowed with the secret intact. The mechanism is worse than "a hole": a URL scheme and a username share a grammar, so `removeprefix(f"{scheme}:")` deleted the password separator before anything looked for it. The `HTTPS:` / `https:` asymmetry is the same bug seen from the other side — `urlparse` lowercases the scheme, so the strip silently no-ops and the colon survives to be caught. My structural claim from last round rested on this function; it was wrong, and I withdraw it.

**F2 confirmed**, and my earlier measurement was the thing at fault: I benchmarked `"%25" * n`, which is many *shallow* escapes and decodes in two passes. The adversarial input is one *deeply nested* escape (`"%" + "25"*d + "40"`), which reproduces your numbers — 64 KB → 1.80 s. The new test's payload took **28 s** before the bound.

**F5 confirmed dead** — identity for every input, across all three call paths. **F6 confirmed** — `https://host:8443/a@b` was falsely refused.

### What parse-or-refuse accepts

Four shapes, each positively matched:

| shape | example |
|---|---|
| URL whose authority resolves | `https://host/org/repo.git`, `ssh://git@host/x` |
| scp-style SSH | `git@github.com:org/repo.git` |
| absolute local path | `/srv/repos/x.git` |
| `file:` with no authority and no `@` | `file:///srv/repos/x.git` |

Everything else is refused.

### What it now refuses that it previously accepted

All of these previously reached `.git/config` with the secret intact:

- `oauth2:TOKEN@gitlab.com:org/repo.git` — GitLab's documented form, scheme dropped
- `x-access-token:TOKEN@github.com/org/repo.git` — GitHub's, scheme dropped
- `user:pw@github.com:org/repo.git`, `user%3Apw@github.com:org/repo.git`
- `https:tok@host/x`, `git:tok@host/x`, `ssh:tok@host/x` — scheme without `//`

And these were already refused but are now refused by parse rather than by surgery: `http:///u:p@h/x`, `//u:p@h/x`, percent- and double-encoded authorities.

**One shape is newly *accepted*** — `https://host:8443/a@b` (F6). A real parse reduces false refusals, as you predicted.

**Two additions I made beyond your prescription**, both because the strict rule refused something legitimate:

1. **Absolute local paths.** Three existing tests use a bare `/tmp/.../remote.git` as `repo_url` — a valid local git remote with no scheme and no authority. Strict parse-or-refuse rejected it. Accepted via a positive match (leading `/`, not `//`, so it cannot be protocol-relative).
2. **`file:` URLs**, which have an empty authority by design. Nothing in the repo uses them, but refusing a credential-free standard form is the false-refusal failure mode you asked me to avoid.

Both are shape matches, not content exemptions.

### `ghp_TOKEN@github.com:org/repo.git` — the residual, stated plainly

Your table lists this as a leak, and your prescription accepts it: it *is* `user@host:path` with no colon before the `@`. I have kept it accepted, because scp syntax is SSH, SSH has no URL password field, and the string is structurally a username. There is no shape-based way to tell a username from a pasted token — only a content-based one, which is the approach we are abandoning. If you want it refused, that is a decision to drop scp-style support, not a detection improvement.

### The operational note, handled

The refusal message now says an existing checkout may still hold a remote written before this check, and tells the operator to delete the checkout directory. I did not try to auto-clean: when the URL is refused there is no safe replacement to write, and clearing `origin` would break a repo that is otherwise fine.

### F3, F4, F7 — noted, and one acted on

**F3** you are right, and I have removed the reasoning's dependency on it: `_git_auth_env` no longer rebuilds a credential-bearing URL into `GIT_CONFIG_KEY_0` at all, since that path went away with `_authenticated_repo_url`. The remaining branch restores a query-string secret and is now guarded by an explicit check that the authority carries no userinfo. **F4** and **F7** left as-is: `https://user:s3cr3t` is genuinely `host:port` to every parser, and F7 is a deliberate asymmetry — the redactor drops SSH usernames from *displayed* text while the guard keeps them in the *persisted* remote, because one is cosmetic and the other has to stay a working URL.

### Verification

- All ten shapes from six rounds refused or sanitized; every supported form still accepted (table above pinned by two parametrized tests).
- Cap: 256 KB token now 0.003 s, was 28 s.
- No regressions: 0 of 20 realistic stderr lines altered, refspec syntax intact, rounds 1–5 shapes still closed.
- **655 tests pass**; `pre-commit` and `tach` clean. Eleven new tests fail against the previous commit.

PR #1717, opened 2026-07-29
Give indexing_settings.json one state type, loader and writer

basnijholt:
## The defect: one file, two schemas, three parsers

Every knowledge base keeps exactly one state file at
`<storage_root>/knowledge_db/<storage_key>/indexing_settings.json`. Both of these resolved to it, by construction from the same `storage_root`, `base_id` and resolved `knowledge_path`:

| | `manager._PersistedIndexState` | `registry.PublishedIndexState` |
| --- | --- | --- |
| Fields | 7 | the same 7 plus `refresh_job`, `reason`, `last_error`, `updated_at`, `last_refresh_at`, `consecutive_refresh_failures` |
| Statuses | `resetting`, `indexing`, `complete` | those plus `failed` |
| Parse | `parse_index_metadata_fields(..., require_complete_fields_for_all_statuses=True)` — rejects any record missing `indexed_count`, `source_signature` or a collection, whatever its status | the same helper, lenient — only a `complete` record must prove itself |
| Path | built inline in `__post_init__` | `published_index_metadata_path` |

A third reader, `manager._published_collection_for_cleanup`, bypassed both parsers and read the raw JSON payload. Its docstring said why: the strict parser "rejects metadata that is merely incomplete, which would silently drop that proof" — and the proof it dropped decides whether a live published collection is deleted as a stale candidate.

### The field-dropping sequence

1. `_save_candidate_publish_metadata` called `write_index_metadata_payload(...)` with seven keyword arguments.
2. That writer wrote exactly the keys it was given, so the six refresh-job fields were **not written**.
3. Reloading the file produced dataclass defaults for all six: both timestamps became `None` and `consecutive_refresh_failures` reset to `0`.
4. `registry._state_with_refresh_fields` then load-modify-wrote them back on the next registry call — two uncoordinated writers to one file, and the drop was invisible only because the refresh runner happened to overwrite the same six fields immediately afterwards.

## What changed

`PublishedIndexState` now lives in `knowledge/index_metadata.py` with one loader and one writer. That module was already the leaf both `manager` and `registry` are allowed to import, and it already owned this file's JSON encoding and `write_json_atomic` (shared with `candidate_checkpoint`); a separate `index_state.py` would have been a third file for one document's schema, either duplicating those primitives or importing them straight back. `registry` imports and uses all three names for its own work, and `refresh_runner`, `status` and `refresh_scheduler` keep reaching them *through* `registry` — a pass-through for those three, kept deliberately so this PR does not churn files other in-flight work is editing.

Deleted: `_PersistedIndexState`, `_load_persisted_index_state`, `_save_persisted_index_state`, the `_INDEXING_STATUS_*` constants, `parse_index_metadata_fields` and its `require_complete_fields_for_all_statuses` flag, `write_index_metadata_payload`, `load_index_metadata_payload`, and `registry._coerce_refresh_job`.

**The writer takes a whole state.** The guarantee this buys is narrower than "no field can be dropped", and worth stating precisely: *the writer cannot drop a field the caller set*. The **type does not enforce completeness** — dataclass defaults mean `PublishedIndexState(settings=..., status=...)` compiles and the writer will happily persist it. Publication is safe because `state_for_publication` states every field and the loader refuses a `complete` record that proves nothing, not because anything at the type level prevents a partial write. Eleven of the thirteen fields keep defaults — they must, so legacy records parse. What is gone is the old failure mode where a writer silently discarded fields the caller never mentioned because they were not in its signature.

Both publish paths now go through one `state_for_publication` factory in `index_metadata.py`, which states all thirteen: the seven publication fields, plus the refresh-job fields resolved to what a publication means — `refresh_job="idle"`, `reason=None`, `last_error=None`, failure streak `0`, both timestamps stamped. That factory also fixes a live instance of the remaining gap: the file-mode publish in `refresh_runner.py` built the same state inline and omitted `consecutive_refresh_failures`, taking the dataclass default — correct today only by coincidence.

### Why publication resolves the refresh fields instead of preserving them

**This is what keeps behaviour identical.** On `main` the publish write emits none of those four keys, so reloading the file yields exactly `refresh_job="idle"`, `reason=None`, `last_error=None`, `consecutive_refresh_failures=0`. Stating those same values explicitly reproduces `main`'s effective state bit for bit. *Preserving* them would have been the behaviour change, not the fix.

And it would have been a bad one. A publish runs while `refresh_job="running"`/`reason="refreshing"` is on disk (`_save_refreshing_state` writes it before the manager starts). Preserving that makes the just-published `complete` record read as `refreshing` → availability `STALE` → the `READY` guard in `_reconcile_cancelled_refresh` (`refresh_runner.py:967`) fails → a fully-published index gets marked stale. `test_cancelled_publish_metadata_save_keeps_published_candidate_collection` asserts the opposite today. Making preservation safe needs reconciliation to tell "a new publication landed" from "the refresh moved" by comparing against `initial_state` — exactly PR 2's `RefreshOutcome` territory.

The only real delta versus `main` is that `updated_at` and `last_refresh_at` are now stamped by the publish write instead of being blanked and refilled a moment later; no reader consults either. `test_publishing_states_every_field_of_the_state_file` pins this and **fails on `main`**, where all four keys are absent from the file after a publish.

### `_published_collection_for_cleanup` loses its escape hatch

Replaced by `_published_state_and_cleanup_safety`, which returns `(state, cleanup_is_safe)` from **one** read:

```python
state = load_published_index_state(self._indexing_settings_path)
return state, state is not None or not self._indexing_settings_path.exists()
```

Safety semantics are unchanged where it matters:

- **File absent** → nothing published, cleanup runs, nothing to protect.
- **File present, no state** (torn JSON, unusable settings, unknown status, or a `complete` record that proves nothing) → `cleanup_is_safe=False`. Cleanup is skipped and the surviving checkpoint is discarded rather than resumed. Never guessed at, never deleted.
- **File present with a state** → whatever collection it names is protected, whatever its status. A collection name is only ever recorded by a publication, so an `indexing` or `failed` record still names the live index — that is the case the raw-payload read existed for, and it is now covered by the one parser.

`_open_candidate_run` consequently reads the file once instead of twice, and its two collection names (`published_collection` from the strict parse, `live_collection` from the raw payload) collapse into one.

## On-disk compatibility: no migration, and none needed

**Decision: the unified parser accepts every payload the previous version wrote, so no existing corpus is re-embedded.** No schema version, no migration step, no shim.

- Fields absent from older files (`refresh_job`, `consecutive_refresh_failures`, the timestamps) already parsed as `idle`/`0`/`None`, and still do — `test_legacy_metadata_without_failure_counter_parses_as_zero` covers this.
- The one validity rule that survives is the registry's: a record that calls itself `complete` must carry `indexed_count`, `source_signature`, and a collection unless it is file-mode. Every writer of a `complete` record has always filled those, so this rejects only corrupt or hand-edited files — and the base is refreshed rather than served from a publication it cannot substantiate (`test_index_metadata_without_source_signature_is_unavailable_and_schedules_refresh`).
- The manager's side becomes strictly *more* accepting: records it used to reject (status `failed`, or a non-`complete` record missing bookkeeping) now parse, which removes spurious forced full reindexes rather than causing any.

The parser is therefore not *fully* non-lossy: an unprovable `complete` record is still no state at all. Making it total would have changed what the registry serves — a behaviour change in availability semantics, not a unification — so it stayed out.

## Deferred follow-up

Moving persistence out of `KnowledgeManager` entirely, so it returns an outcome and `refresh_runner` performs the single write, is deliberately **not** in this PR. It depended on `RefreshOutcome`, which **has now landed on `main` in #1714 and is merged into this branch** — so the follow-up is unblocked, just not done here.

That follow-up is where the two remaining gaps close together: giving the write a single owner is what lets reconciliation know whether a publication landed during the refresh it is reconciling (so publication can preserve the refresh-job fields instead of resolving them), and it is also what makes a lock over the read-modify-write in `_state_with_refresh_fields` meaningful.

## Verification

- `uv run pytest tests/test_knowledge_manager.py tests/test_knowledge_resumable_refresh.py tests/test_knowledge_index_metadata.py tests/test_knowledge_indexing_config.py tests/test_bot_knowledge.py -nauto --no-cov` → **385 passed**
- `uv run pytest tests/api/test_knowledge_api.py -nauto --no-cov` → **62 passed**
- `uv run pre-commit run --all-files` → all hooks pass
- `uv run tach check --dependencies --interfaces` → pass (`index_metadata` now declares its `indexing_config` dependency)

## Review follow-ups (c9a780a)

- **Malformed enum values crashed the loader.** `[] in frozenset({...})` raises `TypeError: unhashable type: 'list'`, so a payload with a JSON array or object in `status` or `refresh_job` escaped a loader every caller treats as total, taking manager initialization and every registry read with it. For `status` this was a regression in this PR (the parser it replaced guarded with `isinstance`); for `refresh_job` it was pre-existing in `registry._coerce_refresh_job`. Both now check for a string first, with array and object cases in the parametrized loader tests.
- **Two candidate-cleanup tests had stopped exercising their target.** Their fixture wrote a `complete` record with no proof of what it published, which the unified loader refuses, so the assertions passed through the "metadata unreadable, skip cleanup" branch rather than through collection-name protection. They now write a `failed` refresh record that still names the live collection, and the cleanup test additionally asserts an abandoned candidate is still reclaimed — an assertion that fails if cleanup is skipped, which is what made the old version vacuous.
- **The pre-candidate compatibility fixture went back to raw JSON.** With one canonical writer that always emits `refresh_job` and `consecutive_refresh_failures`, only a hand-built payload can be the shape an older version left behind; the test now asserts the newer keys really are absent before exercising the loader.

## What this PR does *not* fix

**The lost-update window is untouched.** The opening line of this description is "two uncoordinated writers, one file, no transaction" — what shipped fixes the field-dropping half only. `registry._state_with_refresh_fields` is still load-then-save with no lock, reachable from the file watcher (`watch.py:255`) and the API (`api/knowledge.py:228`), unsynchronised against the manager's publish. A watcher read landing before a publish whose write lands after would `replace()` the fresh publication's `collection`/`indexed_count`/`source_signature` back to the previous values. This is identical on `main` — neither caused nor worsened here — but the file is not transactional now, and the next reader should not assume it is. Closing it belongs with the same follow-up as the persistence move, since that is what gives the write a single owner.

## Round-2 follow-ups (loader totality, duplicated publication state)

- **The loader was still not total, and this PR declared it so.** Two holes, both byte-identical to `main` and therefore pre-existing, but called out because the module docstring and this description assert totality and two of the four holes were already closed here. `"²".isdigit()` is `True` while `int("²")` raises `ValueError`, reaching `indexed_count` and `consecutive_refresh_failures` — now `isdecimal()`, which still accepts `"١٢"` because `int` takes it. And `except (OSError, json.JSONDecodeError)` missed `UnicodeDecodeError`, a `ValueError`, so invalid UTF-8 bytes raised out — now `except (OSError, ValueError)`. Both matter because `manager.py` and every registry read treat the loader as total, so an exception takes manager construction down instead of degrading to "no state, refresh the base". Parametrized regression coverage added for both; verified failing against the old bodies.
- **Each enumeration was declared twice** — a `Literal` plus a hand-written `frozenset`. Now `frozenset(get_args(...))`, matching the existing idiom at `custom_tools/matrix_voice_message.py:33`. Adding a status can no longer half-land.
- **`published_collections` was vestigial.** The set existed because there were two names to hold (strict parse plus raw payload); collapsing to one is the headline of this PR, so the container went too.
- **Three rewritten fixtures used an unreachable shape.** `status="failed"` with a non-null `collection` cannot be produced: only the two publish paths set `collection` and both write `"complete"`, and every registry marker either `replace()`s the loaded state or writes `collection=None`. `status="indexing"` with a collection **is** a real historical shape (`b41727cdb` wrote exactly that) and is the legacy record the collection-protection rule exists to serve. Since these were rewritten on fidelity grounds, they now use the faithful shape.

## Round-3 follow-up (the loader fix was necessary but not sufficient)

`isdecimal()` closed the superscript case but not the class. **An arbitrarily long, perfectly well-formed decimal string passes `isdecimal()` and then makes `int()` raise CPython's integer-conversion-limit `ValueError`** (4300 digits by default), reachable on both `indexed_count` and `consecutive_refresh_failures`. Reproduced: `("1" * 5000).isdecimal()` is `True`, `int(...)` raises.

The general lesson, now written into the helper's docstring: **for this class, catching the failure beats predicating on it.** `isdecimal()` is a claim about characters; `int()` has failure modes that are not about characters at all, so no predicate can stand in for the conversion. The conversion is now wrapped in `try/except ValueError`. `isdecimal()` stays, but only to decide *which shapes count as a number* (keeping `"+1"` and `"1_000"` out) — never to decide whether converting is safe.

Audited every other coercion in the loader for the same predicate-then-convert shape:

| Site | Converts? | Verdict |
| --- | --- | --- |
| `_nonnegative_int` float arm | `int(float)` | Safe — `is_integer()` is false for both infinities and NaN, the only floats `int` refuses, and no finite float exceeds its reach |
| `_load_payload` | `json.loads` | Already `except (OSError, ValueError)`, which also covers a huge integer *literal* in the JSON itself |
| `_optional_str`, `_refresh_job`, status check, `_parse_settings` | none | `isinstance`/membership only; nothing to raise |

### DIFF
--- a/src/mindroom/knowledge/index_metadata.py
+++ b/src/mindroom/knowledge/index_metadata.py
@@ -228,16 +228,35 @@ def _nonnegative_int(value: object) -> int | None:
             return None
         case int() if value >= 0:
             return value
+        # ``is_integer`` is false for both infinities and NaN, which are the
+        # only floats ``int`` refuses, and no finite float can exceed its
+        # reach: this conversion cannot raise.
         case float() if value.is_integer() and value >= 0:
             return int(value)
-        # ``isdecimal`` rather than ``isdigit``: superscripts such as "²" are
-        # digits that ``int`` refuses, so the wider test would raise out of a
-        # loader whose callers expect it to answer for any payload. Decimal
-        # digits outside ASCII, such as "١٢", stay accepted because ``int``
-        # takes them.
-        case str() if value.strip().isdecimal():
-            return int(value.strip())
+        case str():
+            return _nonnegative_int_from_text(value.strip())
     return None
+
+
+def _nonnegative_int_from_text(text: str) -> int | None:
+    """Parse a decimal count, treating anything ``int`` refuses as absent.
+
+    Catching the failure beats predicating on it. No property of the
+    characters decides whether ``int`` accepts them: ``isdigit`` admits
+    superscripts like "²" that it rejects, and even a well-formed all-ASCII
+    decimal string raises once it passes CPython's integer conversion limit,
+    which is not about characters at all. A caller of this loader treats it as
+    total, so one ``except`` covering both -- and whatever else ``int`` grows
+    -- is what makes that true. ``isdecimal`` still gates which shapes count
+    as a number, keeping "+1" and "1_000" out; it just no longer decides
+    whether converting is safe.
+    """
+    if not text.isdecimal():
+        return None
+    try:
+        return int(text)
+    except ValueError:
+        return None
 
 
 def _refresh_job(value: object) -> _RefreshJob:
--- a/src/mindroom/knowledge/index_metadata.py
+++ b/src/mindroom/knowledge/index_metadata.py
@@ -76,7 +76,9 @@ def load_published_index_state(metadata_path: Path) -> PublishedIndexState | Non
         return None
     settings = _parse_settings(payload.get("settings"))
     status = payload.get("status")
-    if settings is None or status not in _PUBLISHED_INDEX_STATUSES:
+    # A JSON array or object decodes to an unhashable value, so membership is
+    # only asked once the value is known to be a string.
+    if settings is None or not isinstance(status, str) or status not in _PUBLISHED_INDEX_STATUSES:
         return None
     state = PublishedIndexState(
         settings=settings,
@@ -189,6 +191,6 @@ def _nonnegative_int(value: object) -> int | None:
 
 
 def _refresh_job(value: object) -> _RefreshJob:
-    if value in _REFRESH_JOBS:
+    if isinstance(value, str) and value in _REFRESH_JOBS:
         return cast("_RefreshJob", value)
     return "idle"
--- a/src/mindroom/knowledge/index_metadata.py
+++ b/src/mindroom/knowledge/index_metadata.py
@@ -18,7 +18,8 @@ import json
 import os
 import uuid
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Literal, cast
+from datetime import UTC, datetime
+from typing import TYPE_CHECKING, Literal, cast, get_args
 
 from mindroom.knowledge.indexing_config import IndexingSettings
 
@@ -30,8 +31,11 @@ if TYPE_CHECKING:
 _PublishedIndexStatus = Literal["resetting", "indexing", "complete", "failed"]
 _RefreshJob = Literal["idle", "pending", "running", "failed"]
 
-_PUBLISHED_INDEX_STATUSES: frozenset[str] = frozenset({"resetting", "indexing", "complete", "failed"})
-_REFRESH_JOBS: frozenset[str] = frozenset({"idle", "pending", "running", "failed"})
+# Derived, never restated: a hand-written copy that drifts from the Literal
+# makes the loader silently refuse every record carrying the new value, or
+# makes the cast that follows the membership test a lie.
+_PUBLISHED_INDEX_STATUSES: frozenset[str] = frozenset(get_args(_PublishedIndexStatus))
+_REFRESH_JOBS: frozenset[str] = frozenset(get_args(_RefreshJob))
 
 
 @dataclass(frozen=True)
@@ -117,6 +121,44 @@ def _records_a_publication(state: PublishedIndexState) -> bool:
     )
 
 
+def state_for_publication(
+    *,
+    settings: IndexingSettings,
+    collection: str | None,
+    indexed_count: int,
+    source_signature: str,
+    published_revision: str | None,
+) -> PublishedIndexState:
+    """Return the whole state one successful publication leaves on disk.
+
+    Both publish paths -- the semantic one that swaps in a candidate
+    collection, and the file-mode one that publishes source metadata with no
+    collection at all -- differ only in those two values, so they share this.
+    Building the state twice is how a field gets left to its default by
+    omission, which is the failure this module exists to make impossible.
+
+    Publication also resolves the refresh job it belongs to: the work that job
+    tracked has just landed, so its reason, error and failure streak are
+    cleared by the same write rather than surviving into a ``complete`` record.
+    """
+    now = datetime.now(tz=UTC).isoformat()
+    return PublishedIndexState(
+        settings=settings,
+        status="complete",
+        collection=collection,
+        last_published_at=now,
+        published_revision=published_revision,
+        indexed_count=indexed_count,
+        source_signature=source_signature,
+        refresh_job="idle",
+        reason=None,
+        last_error=None,
+        updated_at=now,
+        last_refresh_at=now,
+        consecutive_refresh_failures=0,
+    )
+
+
 def save_published_index_state(metadata_path: Path, state: PublishedIndexState) -> None:
     """Atomically persist one whole state, so no field is lost by omission."""
     optional_fields: dict[str, object | None] = {
@@ -155,9 +197,12 @@ def write_json_atomic(path: Path, payload: Mapping[str, object]) -> None:
 
 
 def _load_payload(metadata_path: Path) -> dict[str, object] | None:
+    # ``ValueError`` rather than ``JSONDecodeError``: invalid UTF-8 bytes raise
+    # ``UnicodeDecodeError``, which is a ``ValueError`` and not an ``OSError``,
+    # and every caller of the loader treats an unreadable file as "no state".
     try:
         payload = json.loads(metadata_path.read_text(encoding="utf-8")) if metadata_path.exists() else None
-    except (OSError, json.JSONDecodeError):
+    except (OSError, ValueError):
         return None
     return dict(payload) if isinstance(payload, dict) else None
 
@@ -185,7 +230,12 @@ def _nonnegative_int(value: object) -> int | None:
             return value
         case float() if value.is_integer() and value >= 0:
             return int(value)
-        case str() if value.strip().isdigit():
+        # ``isdecimal`` rather than ``isdigit``: superscripts such as "²" are
+        # digits that ``int`` refuses, so the wider test would raise out of a
+        # loader whose callers expect it to answer for any payload. Decimal
+        # digits outside ASCII, such as "١٢", stay accepted because ``int``
+        # takes them.
+        case str() if value.strip().isdecimal():
             return int(value.strip())
     return None
 
--- a/src/mindroom/knowledge/index_metadata.py
+++ b/src/mindroom/knowledge/index_metadata.py
@@ -1,81 +1,142 @@
-"""Published knowledge index metadata JSON helpers."""
+"""The one persisted knowledge index state document.
+
+Each knowledge base keeps exactly one state file,
+``<storage_root>/knowledge_db/<storage_key>/indexing_settings.json``, and this
+module owns its whole schema: one dataclass, one loader, one writer.
+
+Two collaborators write that file. The indexing manager records a publication
+(which collection is live, what it contains, which revision produced it), and
+the registry records the refresh job around it (queued, running, failed, and
+the error and timestamps that go with it). Both go through
+``save_published_index_state`` with a *whole* state, so neither can reset the
+other's fields merely by not knowing about them.
+"""
 
 from __future__ import annotations
 
 import json
 import os
 import uuid
-from typing import TYPE_CHECKING
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Literal, cast
+
+from mindroom.knowledge.indexing_config import IndexingSettings
 
 if TYPE_CHECKING:
-    from collections.abc import Container, Mapping
+    from collections.abc import Mapping
     from pathlib import Path
 
 
-_IndexMetadataFields = tuple[dict[str, str], str, str | None, str | None, str | None, int | None, str | None]
+_PublishedIndexStatus = Literal["resetting", "indexing", "complete", "failed"]
+_RefreshJob = Literal["idle", "pending", "running", "failed"]
+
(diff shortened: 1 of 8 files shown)


basnijholt:
Reviewed all three Qodo findings against the code. All three are genuine; fixed in c9a780a.

**3. Malformed enums crash loading — real bug, fixed.** Verified: `[] in frozenset({"a"})` raises `TypeError: unhashable type: 'list'`, so a payload with `"status": []` escaped a loader every caller treats as total. For `status` this was a regression in this PR (the parser it replaced guarded with `isinstance(raw_status, str)`); for `refresh_job` it was a pre-existing defect in `registry._coerce_refresh_job`, carried over faithfully. Both now check `isinstance(value, str)` first, and the loader tests parametrize array and object values for each.

**2. Incomplete test exercises corruption path — correct, fixed.** The fixture wrote a `complete` record with no `indexed_count`/`source_signature`, which the unified loader refuses, so the assertion passed through the "metadata unreadable, skip cleanup entirely" branch rather than through collection-name protection. The test now writes a `failed` refresh record that still names the live collection — the realistic shape after a publish followed by a failed refresh, and exactly the case the deleted raw-payload read existed for — asserts the record loads, and additionally asserts an abandoned candidate collection is still reclaimed. That last assertion fails if cleanup is skipped, which is what made the old test vacuous. Renamed accordingly. The two sibling tests using the same fixture (`test_checkpoint_naming_a_published_collection_is_rejected_after_a_failed_refresh`, `test_incompatible_checkpoint_never_deletes_a_published_collection`) had the same problem and got the same fixture.

**1. Legacy fixture writes current fields — correct, fixed.** With one canonical writer that always emits `refresh_job` and `consecutive_refresh_failures`, only raw JSON can produce the shape an older version left behind. The pre-candidate compatibility test now builds the payload with `write_json_atomic` and asserts the newer keys are absent before exercising the loader.

basnijholt:
Round-2: verified all five against the code, all five correct, all five fixed in f550748. PR body corrected for the three framing caveats.

**F1 — loader totality (both holes reproduced).** `"²".isdigit()` is `True`, `int("²")` raises `ValueError`; `UnicodeDecodeError` is a `ValueError` and not an `OSError`, so `except (OSError, json.JSONDecodeError)` let it out. Fixed with `isdecimal()` (confirmed `"١٢".isdecimal()` is `True` and `int("١٢") == 12`, so the accepted set only loses digits `int` refuses) and `except (OSError, ValueError)`. Parametrized coverage for both; verified the new tests fail against the old bodies (2 failed / 30 passed with the two words reverted). Agreed on the reasoning: pre-existing is not a defence when the same PR asserts totality and closes two of four.

**F2 — duplicated publication state.** Confirmed `refresh_runner._publish_file_mode_source_metadata` builds the same thirteen-field state inline and omits `consecutive_refresh_failures`. Both paths now call `state_for_publication(...)` in `index_metadata.py`; only `collection` and `indexed_count` differ, and `manager._state_for_publication` is gone. Tach updated in the same commit (`index_metadata` visibility gains `refresh_runner`, which gains the dependency) rather than routing the call through `registry` as a pass-through.

**F3 — enumerations declared twice.** Now `frozenset(get_args(_PublishedIndexStatus))` / `frozenset(get_args(_RefreshJob))`, matching `custom_tools/matrix_voice_message.py:33`.

**F4 — vestigial container.** Gone; `checkpoint.collection == published_collection` and `preserved = {checkpoint.collection, published_collection} - {None}`.

**F5 — unreachable fixture shape.** Confirmed by enumerating the writers: `_state_with_refresh_fields` only reaches `status="failed"` when there is no current state, and that branch writes `collection=None`; otherwise it `replace()`s and keeps the loaded status. So `"failed"` + collection cannot occur. `b41727cdb` does write `_INDEXING_STATUS_INDEXING` together with `collection=live_collection_name`, so `"indexing"` is the faithful shape — switched, with the tests renamed to match.

**On the framing.** Your correction is right and the PR body now leads with it: on `main` the publish write emits none of those four keys, so a reload yields exactly `idle`/`None`/`None`/`0` — stating them explicitly reproduces `main` bit for bit, and preservation would have been the behaviour change. I have also dropped the overstated crash-window claim (the index stays queryable, reads `STALE`, self-heals) and narrowed the guarantee to what it actually is: *the writer cannot drop a field the caller set*. Eleven fields still have defaults, F2 was a live instance of that gap, and the body says so. Added a "What this PR does not fix" section for the untouched lost-update window in `_state_with_refresh_fields`, and dropped the "re-exports nothing" claim — for `refresh_runner`, `status` and `refresh_scheduler` it is a pass-through.

Local: 464 passed across the six knowledge suites, pre-commit clean, tach clean.

basnijholt:
Round-3: confirmed and fixed in d4a1e3b. The round-1 `isdecimal()` fix was necessary but not sufficient — accepting that fully.

**Reproduced:** `("1" * 5000).isdecimal()` is `True`; `int(...)` raises `ValueError: Exceeds the limit (4300 digits) for integer string conversion`. Reachable on `indexed_count` and `consecutive_refresh_failures`, both parsed by the same helper.

The conversion is now wrapped in `try/except ValueError`. `isdecimal()` stays, but its job is narrowed to deciding *which shapes count as a number* (keeping `"+1"` and `"1_000"` out) — never to deciding whether converting is safe. The reasoning is written into the helper's docstring so the next person does not re-derive it: `isdecimal` is a claim about characters, `int` has failure modes that are not about characters, so no predicate can stand in for the conversion. Verified the new parametrized case (`past_the_integer_conversion_limit`, asserted on both fields) raises without the `except` and passes with it.

**Audited every other coercion in the loader for the same predicate-then-convert shape:**

| Site | Converts? | Verdict |
| --- | --- | --- |
| `_nonnegative_int` float arm | `int(float)` | Safe — `is_integer()` is false for both infinities and NaN, the only floats `int` refuses, and no finite float exceeds its reach. Confirmed `float("inf").is_integer()` and `float("nan").is_integer()` are both `False`, and `int(1e308)` succeeds |
| `_load_payload` | `json.loads` | Already `except (OSError, ValueError)` from the previous round; also covers a huge integer *literal* in the JSON itself, which `json.loads` itself refuses with `ValueError` |
| `_optional_str`, `_refresh_job`, the status check, `_parse_settings` | none | `isinstance`/membership only; nothing to raise |

**PR body updated** with the finding, the audit table, and the caveat you and the first reviewer both reached independently: the type does **not** make partial persistence impossible. Dataclass defaults mean `PublishedIndexState(settings=..., status=...)` compiles and the writer persists it. Publication safety comes from `state_for_publication` stating every field plus the loader's `complete` invariant, not from type-level enforcement.

Local after the fix: 465 passed across the six knowledge suites, pre-commit clean, tach clean.

PR #1718, opened 2026-07-29
Split knowledge refresh policy from its scheduling effect

basnijholt:
## Why

Stop the knowledge read path from welding refresh *policy* to refresh *scheduling*, so cooldowns, Git poll intervals, and failed-refresh retry fingerprints can be exercised without a live scheduler. Deliberately avoids `manager.py` (sibling PRs own it).

> **Two claims in an earlier revision of this description were wrong and are corrected below** — see *Corrections*. Both had already been repeated onward before they were caught.

## The split

`_schedule_refresh_for_availability` used to run while *resolving* an agent's knowledge for a query: it spawned subprocess refreshes and mutated a module-global cooldown dict from inside availability resolution, with the scheduler call appearing in three separate branches.

Policy now lives in `knowledge/refresh_policy.py` as four steps the effect caller drives in order:

| Step | Purity | Role |
|---|---|---|
| `refresh_trigger(...) -> _RefreshTrigger \| None` | pure, cheap | Does this availability warrant a refresh, and what does the turn report while one runs? `None` = do nothing (the common READY case). |
| *caller* | effect | Probes the scheduler — only reached when a trigger exists. |
| `refresh_cooldown_key(...) -> RefreshCooldownKey` | **reads the filesystem** | Builds the throttle key. Deferred until the probe rules out an in-flight refresh. |
| `cooldown_elapsed(...) -> bool` | pure | Has this key's window expired? |

```python
@dataclass(frozen=True)
class _RefreshTrigger:
    availability_while_refreshing: KnowledgeAvailability  # READY -> STALE; others unchanged
    cooldown_seconds: float
```

`knowledge/utils.py` holds the effect half and now calls `schedule_refresh` in **one** place instead of three.

### One predicate, not two

An earlier revision of this PR had `decide_refresh` plus a `scheduler_probe_required` guard, and claimed drift between them was "merely wasteful, never wrong." That holds in one direction only. Adding a new reason a READY index should refresh — embedder signature change, source mtime, a new binding flag — to the decision alone would leave the guard returning `False`, the caller returning early, and the index stale permanently with **no test failing**. `refresh_trigger` is now the single predicate.

### Both clocks injected

`monotonic_now` (cooldown windows) and `wall_now` (age of a persisted timestamp) are both parameters. Previously only the first was, so `_git_poll_due` could not be evaluated at a chosen instant and tests had to bend around wall time. The poll-interval boundary is now pinned at exactly 59 / 60 / 61 seconds against a 60-second interval.

### Not a pure module, and no longer described as one

Fingerprinting a REFRESH_FAILED retry reads a process-global memo of credentials managers and `stat()`s the credential file. That blocking read predates this split and is preserved deliberately — it is what lets a rotated credential retry immediately instead of waiting out the cooldown. It is isolated in `refresh_cooldown_key`, documented there and in the module docstring, and still skipped when a refresh is already in flight.

### Sampling order

The cooldown clock had moved ahead of the scheduler probe; it is now sampled after it, matching `main`. To be precise about the impact, since it was initially reported to me as a live throttle regression: **cadence was provably unchanged.** With arrivals `a`, `b` and probe durations `d_A`, `d_B`, the old code fires when `b − a >= C + (d_A − d_B)` and the new when `b − a >= C`. Probe latency appears on both sides and cancels; under constant latency the two are identical at any latency, and under variable latency the divergence is zero-mean jitter, not a systematic shortening. `is_refreshing` is a target resolution plus a dict lookup, so the asymmetry needed to matter is not reachable. The order is restored because the stamp should record *when the refresh was actually scheduled* — not because a throttle was broken.

## What is not here: `bounded_map.py`

An earlier revision added a shared `BoundedMap` behind the three bounded knowledge caches. **It has been dropped and the three inline prunes restored.** `registry.py` is byte-identical to `main`; `refresh_runner.py` retains only a comment.

It did not earn its place:

- The three prunes shared no code beyond `excess = len(x) - capacity`. Selection, ordering, and protection all differed.
- Each of its three knobs (`tracked`, `pinned`, `eviction_order`) had **exactly one** caller.
- A fourth cache, `_active_refresh_counts`, had to opt out with a comment — it is a refcount map, and capping it would report an in-flight refresh as idle.
- Per-cache semantics moved from three prune functions into three predicates *plus* a generic engine: same piece count, one more layer.
- Its motivating footgun — an insert site forgetting to prune — was audited across every insert site on `main` and was **never live**.
- It also cut against this PR's own thesis. A map that silently mutates itself on insert is an implicit effect in a change whose point is that decisions are pure and callers perform effects.

`_active_refresh_counts` keeps the comment explaining why it is uncapped; that part was worth keeping.

## `status.mark_knowledge_source_changed_async` — kept

Flagged to me as a dead pass-through to delete. It is not dead: `src/mindroom/api/knowledge.py` imports it from `status`, and Tach does not permit `mindroom.api.knowledge` to depend on `mindroom.knowledge.registry`, so the facade is its only legal route. Deleting it broke the suite immediately. Its docstring now names the caller it exists for.

## Corrections

Two rationales in an earlier revision of this description were false.

**1. There was no `monkeypatch` trap.** I wrote that moving cap literals onto the maps "removes the trap the old `monkeypatch.setattr(module, "_MAX_REFRESH_LOCKS", 1)` fell into." Wrong: `_prune_refresh_locks_locked` read the module global **at call time** (`target_size = max(_MAX_REFRESH_LOCKS - reserve_slots, 0)`), so that patch always worked. The test changed only because I had deleted the constant. Moot now — the constants are back.

**2. "All carried over verbatim" was overstated.** Every numeric value was identical, but the cooldown *epoch* moved from post-probe to pre-probe, which a value-only comparison cannot catch. Addressed above.

## Throttles

Unchanged: 300s retry cooldown, Git poll-interval derivation (`max(poll_interval, 1.0)`), READY on-access cooldown, and the 128 / 512 / 512 caps. The `poll_interval_seconds == 0` asymmetry between the two cooldown helpers is a verbatim move and unreachable anyway — `Field(default=300, ge=5)` rejects `0`.

## Incidental fix

`main`'s `knowledge/utils.py` imported `mindroom.credentials` at runtime without declaring it in Tach. The move to `refresh_policy` declares it, closing a pre-existing gap. (Note that Tach does not govern that edge — `mindroom.credentials` has no `[[modules]]` entry — so `tach check` passed either way; the declaration is for consistency with the ~10 other modules that list it.)

## Deferred

`KnowledgeRuntimeState` (hanging the caches off an orchestrator-owned object) is not in this PR; it would cross `manager.py` and the `refresh_runner.py` ranges sibling PRs own.

## Tests

`tests/test_knowledge_refresh_policy.py` — no scheduler, no event loop, no process-global cooldown state. Covers the trigger short-circuit, READY-to-STALE downgrade, exact poll-interval and cooldown boundaries, Git interval overriding the default throttle, credential-rotation retry, and per-availability cooldown key separation.


basnijholt:
Reviewed both bot findings against the code. One accepted with a corrected rationale, one rejected.

### 1. Declare `mindroom.credentials` in `refresh_policy`'s Tach `depends_on` — accepted, but the stated reason is wrong

Fixed in 2353da1f2.

The recommendation is right, but not for the reason given. Qodo's claim was:

> The repository's enforced `tach check` will reject this undeclared runtime edge.

It does not. `mindroom.credentials` has no `[[modules]] path = ...` entry, so Tach does not govern that edge. I verified both directions locally — `uv run tach check --dependencies --interfaces` reports `✅ All modules validated!` with and without the dependency listed — and the `check` job passed on the first commit of this PR, before the change.

What is actually true is that roughly ten other modules that import `mindroom.credentials` at runtime do list it, so omitting it is inconsistent. The omission is also inherited rather than new: `knowledge/utils.py` held this exact import at module scope and did not declare it either. Declaring it makes the boundary match the import, so it is in.

### 2. Align `poll_interval_seconds == 0` handling between the two cooldown helpers — rejected

CodeRabbit hedged this correctly ("If this is a faithful move of the old behavior, ignore"), and it is one. `_refresh_on_access_cooldown_seconds` is byte-identical to the pre-PR `_schedule_refresh_on_access_cooldown_seconds`, `or` included:

```python
# aa78c7b13:src/mindroom/knowledge/utils.py:273-278  -- and the current refresh_policy.py
    if config.get_knowledge_base_config(lookup.key.base_id).git is None:
        return _REFRESH_RETRY_COOLDOWN_SECONDS
    poll_interval_seconds = _git_poll_interval_seconds(lookup, config)
    return max(poll_interval_seconds or _REFRESH_RETRY_COOLDOWN_SECONDS, 1.0)
```

The divergence is also unreachable. `KnowledgeGitConfig.poll_interval_seconds` is `Field(default=300, ge=5)`, so Pydantic rejects `0` before it reaches either helper, and `_git_poll_interval_seconds` returns `None` only when `git is None`, which the preceding branch already handled. Both fallback arms are dead in both functions.

Changing it would retune a throttle in a PR whose contract is that no throttle value changes, to fix a case the config schema makes impossible. Leaving as-is deliberately.

basnijholt:
Round-1 findings addressed in e076b84d0. Disposition of each, including two where I diverged from the suggested fix and one where I disagree with a stated mechanism.

### Accepted as stated

**Duplicated READY predicate (finding 2).** Correct, and the direction-asymmetry argument is the right one — drift silently strands an index rather than merely wasting a probe. Fixed, though not with the suggested callable; see below.

**`_RefreshDecision.schedule_refresh` was test-only (finding 3).** Confirmed zero production callers. The dataclass it lived on is gone entirely.

**Impurity (finding 1 / finding B).** Confirmed all three: the credentials-manager memo, the `stat()`, and the un-injected `datetime.now`. Two reviewers reaching this independently settled it. `wall_now` is now injected, which also fixes the untestable poll-interval boundary — covered at exactly 59 / 60 / 61 against a 60-second interval.

**The false `monkeypatch` rationale (finding 5).** Verified and fully conceded: `_prune_refresh_locks_locked` read `_MAX_REFRESH_LOCKS` at call time, so the patch always worked. There was no trap; the test changed only because I deleted the constant. Corrected in the PR description, which now carries an explicit corrections section.

### Accepted, different fix

**Finding 2 — a trigger split rather than a `Callable[[], bool]` probe.** The callable would have moved the probe inside the policy, and with it the point at which the cooldown clock must be sampled. That makes the sampling-order question unfixable from the caller. Instead the caller drives four ordered steps and keeps the probe: `refresh_trigger` (pure) → probe → `refresh_cooldown_key` (the one filesystem step) → `cooldown_elapsed` (pure). Same result — one predicate, `scheduler_probe_required` deleted — and the caller still controls ordering.

**Finding B — the retry fingerprint stays inside policy.** The suggestion was to compute it in the caller and pass it in as a value. I did that for the wall clock but not the fingerprint: deciding *when* a fingerprint is needed is itself policy (only REFRESH_FAILED, only after the probe). Hoisting it to the caller would put that condition in two places, recreating exactly the drift hazard finding 2 asks me to remove. It is instead isolated in one clearly-named function, documented as the blocking step, and still skipped when a refresh is in flight. The purity claim is deleted rather than faked.

### Disagree on mechanism, fixed anyway

**The cooldown epoch (finding A).** The observation is right — the clock moved from post-probe to pre-probe — and I have restored the original order. But the stated mechanism, "probe latency is subtracted from the cooldown," does not hold. With arrivals `a`, `b` and probe durations `d_A`, `d_B`:

- new: stamps `a`, checks at `b` → fires when `b − a >= C`
- old: stamps `a + d_A`, checks at `b + d_B` → fires when `b − a >= C + (d_A − d_B)`

Probe latency appears on both sides and cancels. The divergence is `d_A − d_B`: identical at any *constant* latency, zero-mean jitter otherwise. The reported demonstration needs `d_A ≈ 299s` with `d_B ≈ 0`, and `is_refreshing` is a target resolution plus a dict lookup. Every schedule also re-stamps a fresh epoch, so no error can ratchet into a storm.

So this is corrected because the stamp should record when the refresh was actually scheduled — not because a throttle was broken. I have described it that way in the PR body rather than as a regression fix.

### Dropped `bounded_map.py`

Agreed, and removed. `registry.py` is byte-identical to `main` again. The case against it is sound on every count — one subtraction of shared kernel, one caller per knob, a fourth cache forced to opt out, same piece count with a layer added, and a motivating footgun that the insert-site audit showed was never live. The point that carried most weight is the design one: a map that mutates itself on insert is an implicit effect inside a change whose whole thesis is that decisions are pure and callers perform effects.

Worth recording that the two verifications it did pass — 900 randomized trials and an independent invariant proof of the lock pinning — confirmed it was *correct*. It just was not *warranted*, which is a different bar and the right one here.

Kept the one genuinely useful artifact: the comment on `_active_refresh_counts` explaining why that map is deliberately uncapped.

basnijholt:
Round-2 findings addressed in 969d9e9a6, plus a merge of `main` through #1714 in 5810c37d3. All four confirmed; no production logic changed. Two were verified by mutation rather than by inspection, since both were coverage claims.

### Finding 1 — docstring wrong twice

Confirmed both halves. The steps read probe → sample clock → build key, while `utils.py` does probe → build key → sample clock; as written it would put the credential `stat()` between the sample and the stamp, which the caller's own comment forbids. And the rationale was the refuted one — "sampling first would subtract probe latency" — which this PR's description already retracts.

Worth naming plainly: this is the retracted claim shipped in source instead of a description, where it would have outlived the PR and the retraction both. Renumbered to the real order, with the real reason (the stamp should record when the refresh was scheduled) and an explicit note that cadence does not depend on it.

### Finding 2 — vacuous distinctness assertion

Confirmed. Five keys built from five distinct enum members are distinct on the availability component alone; the assertion held for any settings component. Replaced with per-availability assertions on that component:

| availability | asserted settings |
|---|---|
| READY, INITIALIZING, CONFIG_MISMATCH | `indexing_settings` |
| STALE | `None` |
| REFRESH_FAILED | `(indexing_settings, *fingerprint)`, credential absent from `repr` |

Mutation-verified with the exact refactor described — making CONFIG_MISMATCH drop its settings now fails `test_cooldown_key_carries_indexing_settings[CONFIG_MISMATCH]`.

### Finding 3 — lost stamp-then-prune coverage

Confirmed by mutation before fixing: deleting `_prune_refresh_schedule_bookkeeping()` from the production stamp site passed the entire knowledge suite. The loop now stamps through `_schedule_refresh_for_availability` with a scheduler stub, and the same deletion fails with `assert 552 <= 512`.

This one is worth flagging beyond the fix. I argued in this PR's description that the missing-prune footgun was never live, and used that to justify dropping `BoundedMap` — while the same change removed the test that would have caught it at this site. The audit conclusion stands, but it was resting partly on coverage I had just deleted.

### Finding 4 — double wall-clock sample

Confirmed: `_resolve_base_knowledge` and `_schedule_refresh_for_availability` each called `datetime.now`, so one turn evaluated the poll boundary at two instants. Not a regression — `main` did the same — but injecting the clock specifically to pin that boundary and then reading it twice defeats the point. Hoisted to one sample per resolve, passed down.

### Merge of #1714

Merged clean. Post-merge verification:

- `manager._last_refresh_error` reads in `refresh_runner.py`: none.
- `_record_refresh_error` in `manager.py`: 5.
- #1713's deleted symbols across `knowledge/`: none.
- `registry.py`: still a 0-line diff vs `main`.
- `refresh_runner.py`: still only the 3-line `_active_refresh_counts` comment.
- `manager.py`: 0-line diff — this PR never touches it.
- Both mutations above re-run post-merge and still fail correctly.

On the redaction hazard in #1714's five sites: `manager.py` is untouched here, so this PR neither adds nor removes that exposure. It is real though, and the same shape as the two gaps above — a property that is true today, enforced by nothing. Worth its own guard in whichever PR owns `manager.py`.

PR #1719, opened 2026-07-29
Fix a knowledge redaction crash, and stop credentials reaching .git/config

basnijholt:
## What

Two things:

1. **A crash fix.** `urlparse` raises when a netloc holds a codepoint that NFKC-normalises to a URL delimiter, and `knowledge/redaction.py` calls it unguarded. **No credential is involved.** On current `main`:

```
"fatal: unable to access 'https://exa＠mple.com/repo.git'"   -> ValueError
"fatal: unable to access 'https://ho：st/repo.git'"           -> ValueError
```

So a typo in `repo_url` makes `manager.py` raise *while recording* a Git failure, and `api/knowledge.py`'s `_redacted_last_error` returns 500 for as long as that error stays persisted. This makes the helpers total.

2. **A write gate.** `credential_free_repo_url` only strips userinfo from the authority `urlparse` reports, so a URL whose credentials sit elsewhere — an empty authority, a percent-encoded separator, a scheme-dropped provider form like `oauth2:TOKEN@gitlab.com:org/repo.git` — passed through untouched and was written into the checkout's `.git/config` by `git remote set-url`, where it persisted across syncs. `_persistable_remote_url` parses or refuses instead.

Credentials still transit as a process-local `GIT_CONFIG_*` header. Transit yes, persist never — that asymmetry is the design, and is now enforced.

## What this PR is *not*

Earlier revisions of this branch widened *what* gets redacted, across twelve credential shapes. That work is removed. Two reasons:

- **There is already a second, older redactor.** `src/mindroom/redaction.py` is wired in as a global structlog processor plus tool events, `!config` display, LLM request logging and container logs. It sees every URL this codebase logs, not just the knowledge paths — and it leaks five of those twelve shapes. Hardening the module with the narrow blast radius while the wide one goes unexamined is the wrong order.
- **Those shapes all require a credential in `repo_url`**, which `docs/knowledge.md` forbids and `credentials_service` exists to replace, and they still leak in `mindroom/redaction.py` today. Under this PR's stopping rule — *no worse than what ships, better in enumerated ways* — they were never in scope for this PR alone.

`redact_credentials_in_text` is now byte-equivalent to `main` on every diagnostic probed, including several this branch had been **over**-redacting (`note: see a:b@c`, OCI references like `.../v2/app:1.4@sha256:…`, submodule lines carrying `git@host:path`). Breadth changes belong in a follow-up that covers both redactors.

## Not duplication

`credential_free_repo_url`, `fully_unquoted` and `credential_free_url_identity` are **not** overlap with `mindroom/redaction.py`. They must return a **usable** URL — for `.git/config` and for identity comparison — where `***` is not an available answer. That is a different job with an incompatible failure policy. Only `redact_credentials_in_text` overlaps, and that is the piece the follow-up should consolidate.

## Changes

| | |
|---|---|
| `knowledge/redaction.py` | `_inspectable_url` makes every helper total; `fully_unquoted` and `MAX_REDACTABLE_TOKEN_LENGTH` for the gate; widened `except ValueError` on the Basic-token decode (same crash class) |
| `git_urls.py` | `credential_free_repo_url` returns unchanged instead of raising with the netloc quoted |
| `knowledge/git_source.py` | `_persistable_remote_url` + `_parsed_remote_url` (raising `from None`, since `urlsplit`'s message quotes the password and the refresh subprocess logs with `logger.exception`); length bound; `_git_auth_env` guarded locally rather than by call order |
| `docs/knowledge.md` | accepted `repo_url` forms, the refusal class, and the stale-checkout caveat |

## Verification

- Crash: `main` raises on both NFKC forms; this branch returns a redacted string.
- Diagnostics: **0 of 8** probed outputs differ from `main`.
- Gate: 11 unsafe forms refused with secret-free messages; 12 supported forms accepted, including scp with and without a username, underscore hosts, `file:`, absolute local paths, and `@` in a path.
- `uv run pytest` over the knowledge, redaction, API and import-graph suites — **537 passed**.
- `uv run pre-commit run --all-files` and `uv run tach check --dependencies --interfaces` — clean.

## Follow-up

Consolidating `redact_credentials_in_text` with `mindroom/redaction.py`, and deciding redaction breadth once, for both. #1720 tracks a separate quadratic in the shared secret-scrubbing loop.

## Summary by CodeRabbit

* **Documentation**
  * Expanded Git repository URL guidance to cover supported URL and path formats.
  * Clarified rejected URL patterns and credential-handling requirements.
  * Documented cleanup steps for existing checkouts with credential-bearing remotes.

* **Bug Fixes**
  * Improved handling of malformed URLs and invalid authorization data without raising errors.
  * Prevented unsafe or ambiguous repository URLs and credentials from being persisted.
  * Added safer credential injection for Git operations.

* **Tests**
  * Added regression coverage for URL validation, credential redaction, and malformed input handling.


basnijholt:
## Round 8: plan before code

I reproduced everything, and the aggregate matches: **8 of 11 regress** on my corpus. Both write-path bypasses confirmed live. No disagreement with any finding — the PR is a net regression and I am not defending it.

### The one detail that determines the fix

**Trailing framing is fine; leading framing is what breaks.** `https://u:S@host/x,` redacts correctly; `Endpoint=https://u:S@host/x` does not. Because `_URL_LIKE_TOKEN` is anchored `^`, and the tokenizer folded `Endpoint=` into the same token.

Main handles all of these for one reason: **it searches, it does not anchor.** So the regression is not the fail-closed core — it is that I replaced a search with a tokenize-then-anchor. Your "invert the last gate" is right, and the concrete form is: go back to searching for substructure, keep the fail-closed classifier.

### Why I tokenized, and why I no longer need to

I tokenized to kill the O(n²) scan. I measured which regex shapes actually cause it:

```
pattern                        slashes      letters   letters+colon    mixed
main _URL_PATTERN              0.0002s      0.7341s      0.0090s      0.0020s
PR protocol-relative           6.0253s      0.0000s      0.0000s      0.0000s
cand //-then-consume           0.0001s      0.0000s      0.0000s      0.0000s
cand scheme-guarded            0.0004s      0.0003s      0.0003s      0.0003s
```

The rule is precise: **a pattern that can fail *after* consuming is quadratic, because `re.sub` retries at the next position; a pattern that cannot fail once its anchor matches consumes the run and moves on, so it is linear.** My protocol-relative pattern required an `@` after `[^\s]*` — that is the failure. Main's `_URL_PATTERN` has a latent instance of the same thing (0.73 s on 40 KB of letters — pre-existing, not mine, but I will not reintroduce it).

So I can search *and* stay linear, which removes the reason the tokenizer existed. A negative lookbehind stops the scheme star being retried mid-run, and the protocol-relative pattern drops its required-character tail and lets the callback decide.

### Plan

**1. Invert the gate: locate substructure, classify what was located.**

Unanchored linear searches — `(?<![A-Za-z0-9+.-])scheme://…`, `//…`, transport `scheme:…`, scp, and `user:pass@` — each of which cannot fail after its anchor. The callback keeps the positive-recognition core.

**2. The over-redaction falls out, and one nested case must not.**

Once the URL is *located*, an `@` in the path is a located, credential-free character rather than an unclassifiable token, so `https://host/@scope/pkg.git` returns unchanged. But `https://host/https://u:S@in/x` also has its `@` in the path and **is** a credential, so the classifier fails closed when the path contains `://` or a `:`…`@` pair. That distinction is the whole of findings 2 and 3.

**3. `git@github.com: Permission denied` — my scp change was itself a (b) regression.**

You flagged it and you are right. The discriminator is that a remote has a non-empty path after the colon: require `[^@\s]+` there. `git@github.com:org/x.git` still redacts (the round-7 Finding 2 improvement stands); `git@github.com:` followed by prose stops being touched.

**4. Both write-path bypasses.** Move the `fully_unquoted` encoding comparison *above* the scp early return, and refuse when the path contains a nested `://`. Both are ordering/coverage, not design.

**5. The A/B harness as a test**, built exactly as you specified: baseline pinned to a **commit SHA**, generated corpus over `{schemes} × {encoding depth 0-3} × {framing chars incl. none} × {userinfo forms} × {Git and LFS diagnostic templates}`, asserting both directions, with an explicit named allowlist for intentional improvements. Your loader is the right shape — I will pin the rev and drop the `sys.modules` mutation into a fixture so it cannot leak between tests.

I will state the allowlist explicitly rather than letting it absorb failures. Expected entries: the ten closed shapes, and scp-with-a-real-path.

**6. Also:** timing test rebuilt on many sub-bound tokens (your 500×2 KB and 1 MB slash-token cases); `docs/knowledge.md:281` updated for the refusal class including `./repo.git`, `../repo.git`, `~/repo.git` and the `credentials_service` requirement; PR body's "17 diagnostics" corrected — the list has 9 and omits the regressing shapes, which is exactly how the claim survived.

**7. Declining, with reason:** the quadratic `decoded_basic_values` loop. Byte-identical to main, not a regression, and it is a different mechanism (repeated `str.replace`, not scanning). Folding it in would put an unmeasured change inside the round that has to prove it makes nothing worse. Worth its own issue.

### On the acceptance criterion

Agreed, and it is the part I should have asked for three rounds ago. Enumeration produced eleven shapes and two regressions; "provably no worse than what ships, better in exactly these named ways" is checkable and it terminates. I will make it a test in this PR, and it gates the branch rather than the other way round.

One consequence I want stated up front: with that test in place, **the correct response to a twelfth shape found in the branch is only to fix it if main also closes it.** If main leaks it too, it is a pre-existing bug and belongs in a separate change against main — not another round here.

PR #1724, opened 2026-07-29
Move the Git-source tests into their own file

basnijholt:
`src/mindroom/knowledge/git_source.py` was extracted into its own module in #1716, but its tests stayed in `tests/test_knowledge_manager.py` — 9056 lines, 237 collected tests, whose four longest tests were all Git tests.
The module boundary existed; the test boundary did not.
This is a pure move: no behavior changes, no rewrites, no new tests.

## Test counts

| | Collected |
| --- | --- |
| `test_knowledge_manager.py` before | **237** |
| `test_knowledge_manager.py` after | **220** |
| `test_knowledge_git_source.py` (new) | **17** |
| | 220 + 17 = 237 ✅ |

17 collected from 16 test functions — `test_git_unsupported_scheme_userinfo_is_not_copied_to_git_config_env` is parametrized ×2.

`test_knowledge_manager.py` drops 9056 → 7975 lines.

## What moved

The 16 tests whose subject is the checkout itself, not the index built from it:

- **Sync mechanics** — `_sync_once` unchanged-head fast path, LFS pull after reset, repeated-hydration skip, `_ensure_repository` LFS clone with `GIT_LFS_SKIP_SMUDGE`, `_hydrate_lfs_worktree` ignoring index extension filters, missing-`git lfs` guidance, branch switching on an existing single-branch checkout, and `sync()`'s `GitSyncResult` contract.
- **`_run_git` process handling** — timeout kill/reap, cancellation kill/reap, `index.lock` surfaced without retry, credential redaction in error messages.
- **Credential injection** — the four tests proving secrets reach `git` only through process-local `GIT_CONFIG_*`, and never land in the checkout's own `.git/config` or in published index metadata. That is the contract `git_source.py`'s module docstring leads with.

The `_git_manager` helper moved with them: every one of its call sites was in this set, which is the clearest evidence the boundary is real.

## What deliberately stayed

Tests where Git is the *source fixture* rather than the subject:

- `test_git_noop_refresh_*` (7 tests) and their `_git_noop_config` / `_install_git_sync_results` / `_install_git_revisions` helpers — these assert the manager's no-op fast path and corpus-hash decisions.
- `test_git_pull_that_changes_one_file_only_reindexes_that_file` — the claim is indexing *cost* (`indexed_count == 1`), which is the manager's incremental reindex.
- `test_git_worktree_checkout_file_is_detected_for_sync_listing_and_api_status` — spans `file_listing`, `git_source` and the API status endpoint; three subjects, no clean home.
- The `redaction.py` URL unit tests (`test_redact_url_credentials_*`, `test_credential_free_repo_url_*`, `test_git_url_identity_*`) — their subject is `knowledge/redaction.py`, and `credential_free_url_identity` is also used by `indexing_config.py`, so they are not git_source-exclusive.
- Duplicate-git-root config validation, refresh cooldowns, and Git polling/scheduling — config and scheduler subjects that merely configure a Git base.

## Shared fixtures

The moved tests build a `KnowledgeManager`, so they need the `patch_vector_store` fixture; without it `__post_init__` constructs a real ChromaDb and a real embedder.
Rather than duplicate it, the in-memory Chroma fake (`_Collection`, `_Client`, `_VectorDb`, `_Knowledge`, `_FakeEmbedder`), the fixture, and the `_config` factory move to `tests/knowledge_test_support.py`, whose docstring already reads "Shared fakes for knowledge-index tests".
Both files now import one definition. `_insert_with_real_reader` and `_AutoCreatingKnowledge` stay in the manager file — only its tests use them.

The fixture is requested by name via `pytestmark = pytest.mark.usefixtures("patch_vector_store")` rather than left `autouse=True`. Relocating it meant a module activated it purely by importing it, and that failed badly rather than loudly: deleting the import made 6 tests fail while **11 kept passing against a real ChromaDb**. Naming it turns that into a collection error for all 17. Scope is unchanged — one function-scoped activation per test, in exactly the two modules that had it.

The relocated names keep their leading underscore. Making them public turns every fake method into a "public method" for pydocstyle and demands 21 invented docstrings inside otherwise-untouched code; ruff is content with sharing them as-is.

## Verification

- **Byte-identical move.** A script parsed `origin/main:tests/test_knowledge_manager.py`, extracted all 17 moved definitions by AST span, and compared them to the new file: all identical, no tolerance applied. The only additions to `test_knowledge_manager.py` are the shared-import block, the `pytestmark` line, and three import lines that ruff narrowed because the code needing `Lock`, `ClassVar`, `Iterator` and `Sequence` moved out. Every other line in that file's diff is a deletion.
- **No stale or duplicated references.** Each moved test name and `_git_manager` resolves to exactly one definition across `tests/`; no relocated fake is still defined in the manager file.
- **Fixture scope preserved.** `patch_vector_store` was the *only* fixture defined in the manager test module, and it is function-scoped before and after — no test silently changed what it proves. The five autouse fixtures in `tests/conftest.py` apply to both files equally.
- 351 passed across `test_knowledge_git_source.py`, `test_knowledge_manager.py` and `test_knowledge_resumable_refresh.py` (the third also imports the support module). `uv run pre-commit run --all-files` passes, Tach and module-privacy hooks included. No Tach boundary changed — this touches only `tests/`.

## Did this make the codebase better?

Yes, but partially, and it is worth being precise about the limit.

The win is real: `git_source.py`'s contract — subprocess lifetime, timeout and cancellation, LFS hydration, credential injection — now lives in one 884-line file instead of being scattered through the tail of a 9000-line file, and the security-relevant credential tests are together where someone auditing that module will find them. `_git_manager` having no callers outside this set confirms the split follows a seam that already existed rather than one I imposed.

The limit: `test_knowledge_manager.py` is still 7975 lines and 209 tests, and it still mixes refresh, publish, availability, cooldown, scheduling, API and config-validation concerns. This PR removes one cohesive cluster; it does not make that file navigable. Anyone expecting the split to fix `test_knowledge_manager.py` will be disappointed — several further splits along those lines are each as plausible as this one.

One cost I did not anticipate: because the moved tests construct a `KnowledgeManager`, the shared Chroma fake had to move too, so the PR touches that file's spine and not only its tail. That relocation is defensible on its own (it removes a would-be duplication and lands in the module built for it), but it is the reason the diff is larger than "16 tests moved" suggests. It also surfaced the fixture-activation hazard above, which is a genuine find rather than churn.


basnijholt:
## Review triage

Four findings. One was about code this PR introduced and is now fixed; three target code that is byte-identical to `origin/main` and are out of scope for a pure move.

### Fixed — fragile autouse fixture import (Qodo #2)

Valid, and worth more than "maintainability". I tested the failure mode by deleting the import: **6 tests failed but 11 kept passing**, silently constructing a real ChromaDb and a real embedder instead of the fake. So the hazard is not just that the import looks removable — it's that removing it degrades most of the file from unit to integration behavior without any signal.

Fixed in 83c74fa42 by dropping `autouse=True` and naming the fixture in `pytestmark = pytest.mark.usefixtures("patch_vector_store")` in both modules. Deleting the import now produces 17 collection errors instead of 11 quiet passes. Scope is unchanged: one function-scoped activation per test, in exactly the two modules that had it.

I did **not** take the suggested remedy of moving it to `tests/conftest.py`. That would make it autouse for the entire suite, patching `mindroom.knowledge.manager.ChromaDb`, `...collections.ChromaDb` and `create_configured_embedder` for hundreds of unrelated tests — a much larger behavior change than the problem warrants. `pytest_plugins` has the same global effect.

### Out of scope — pre-existing code, moved verbatim

This PR relocates 17 definitions with zero content edits; a script verifies each one is byte-identical to `origin/main:tests/test_knowledge_manager.py` by AST span, with no tolerance applied. That check is what makes the diff safe to review, so changing assertions here would remove the evidence that nothing changed behaviorally.

1. **">3 patches per test" (Qodo #1)** — pre-existing patch counts, unchanged by this PR. The repo's 3-patch guideline is about testing through the right seam, and these tests are already at the owning seam: they drive `GitKnowledgeSource._sync_once` while stubbing `_run_git`/`_rev_parse` because there is no real remote. Reducing the count means redesigning how the Git subprocess boundary is faked — a separate change with its own risk, not a side effect of moving files.

2. **`_Collection.get`/`delete` apply only the first `where` condition** — pre-existing fake behavior, moved unchanged. The observation is correct and the reasoning is good (this file does guard other unsupported shapes), but production only issues single-field `where`, so nothing is currently mis-tested. Worth a follow-up on the fake itself.

3. **`getattr()` on a coroutine in `_fake_wait_for`, and the two `hasattr()` assertions in `test_git_source_sync_does_not_mutate_index_directly`** — both pre-existing, both moved unchanged. The `hasattr` critique is factually right: those two lines assert the static shape of `KnowledgeManager` and would pass with an empty test body; the real claim is carried by the `GitSyncResult`/`last_synced_head` assertions below them. That makes them dead weight rather than wrong, and deleting them is a behavior-neutral cleanup that belongs in a PR that is allowed to edit test bodies.

Happy to take 2 and 3 as a small follow-up against these tests now that they have their own file — which is part of the point of giving them one.

PR #1726, opened 2026-07-29
Pin knowledge reader selection and chunk boundaries with tests

basnijholt:
## What this is now

**Tests only. There is no production diff.** `git diff origin/main -- src/ tach.toml CLAUDE.md` is empty.

Reader selection and chunking policy decide what a knowledge base's vector store contains, and nothing asserted them. **Seventeen mutations of that code passed the entire existing suite** — including one that corrupts a process-wide shared cache. This PR pins them.

| File | Cases | What it holds |
| --- | --- | --- |
| `tests/test_knowledge_readers.py` | 19 | which reader answers for a file, and the chunk boundaries it produces |
| `tests/test_knowledge_resumable_refresh.py` | 2 | the embedding prefetch's reader gate and its byte budget at the exact-fit boundary |

## Withdrawn: this began as an extraction, and that part is reverted

Earlier revisions moved the reader concern into a new `src/mindroom/knowledge/readers.py`. That is gone. The accounting that ended it:

```
manager.py   +31  -95     (net -64)
readers.py  +170   -0
tach.toml     +6   -1
            ------------
net          +111 production lines, behavior verified identical
```

+111 net production lines to make a 1965-line file 3.7% shorter, fixing no bug, and introducing a failure mode that did not exist before: chunking went from a method reading `self.config` to a `chunking=` parameter that four call sites had to supply correctly. It also carried an encoding guard no call path could reach.

Two rationales offered for it in earlier revisions were wrong and are withdrawn:

- **"It removes active duplication."** It did not. `build_reader` and the prefetch gate looked like one question asked twice, and the claimed consequence — prefetching chunks no insert produces — cannot occur, because both paths take their reader from an identical `build_reader` call.
- **"These tests were not writable without it."** Also wrong. `KnowledgeManager` already has the behavior; the tests only needed the behavior, not a new module.

Once the second claim fell, the only argument left for merging the extraction was the work already spent on it. So the extraction goes and the coverage stays.

## The one real decision: why these tests use private methods

They call `_build_reader`, `_configure_text_reader` and `_chunking_strategy` instead of driving `reindex_all`. That is a deliberate exception to this repository's preference for public seams, and it is forced by the hazard the most important test guards.

`ReaderFactory` caches **one reader instance per extension**, and `_build_reader` reconfigures a copy of it. Delete that `deepcopy` and **both existing end-to-end suites still pass** — measured, not assumed. The reason is structural: within a single refresh, every `_build_reader` call re-applies that manager's own chunking before the reader is used, so a sequential refresh stays internally consistent. Corruption requires two bases' refreshes to *interleave* — reachable in production, since the source-root lock is per `KnowledgeSourceRoot` and prefetch reads run under `asyncio.to_thread`, but not stageable in a deterministic test.

So an end-to-end test cannot hold this invariant. Observing the constructed reader can. Under the mutant, two bases at 137/11 and 2000/0 receive the *same object*, base A indexes at B's chunk size, and the factory cache serves B's size for the rest of the process.

The boundary tests use the same seam for consistency. The end-to-end boundary assertion already exists on main (`test_overlapping_chunks_are_batch_prefetched_and_published`) and is not duplicated here.

File placement: kept as its own concern-scoped file rather than folded into `test_knowledge_manager.py`, which is already ~8,000 lines; adding a distinct concern there makes it harder to navigate, not easier.

## Mutation results — 17 tried, 17 fail

All against **unmodified** `manager.py`.

| mutation | real effect |
|---|---|
| `deepcopy` dropped, text branch | two bases share the factory's cached reader; one re-chunks the other's corpus and poisons the cache process-wide |
| `deepcopy` dropped, JSON branch | the shared chunker is aliased rather than copied (inert today) |
| `_chunking_strategy` hardcoded | the base's authored config is ignored |
| `chunking_strategy` never assigned | text files revert to the factory's 5000/0 |
| `reader.chunk_size` never set | reader and strategy disagree on the size |
| supplied strategy's `overlap` zeroed | every boundary in the corpus moves |
| `min_chunk_fill_ratio` 0.5 → 0.25 | boundary placement shifts on whitespace-poor text |
| text gate narrowed to `TextReader` | every `.md` reverts to factory chunking |
| `_FallbackAwareJSONReader` → plain `JSONReader` | the malformed-source fallback stops existing |
| JSON reader copied with `chunk=False` | a JSON element over 5000 characters becomes one oversized vector |
| JSON `read` drops `name=` | every chunk of a file is stored under the wrong `name` metadata, which Chroma queries by |
| JSON tagging made unconditional | the whole corpus diverts to the text fallback |
| `line`/`column` swapped in the parse error | log-only, but the old sample had both equal to 3 and observed nothing |
| fallback reader's chunk flag forced off | one long malformed file becomes a single oversized vector |
| fallback `read` drops `name=` | same misfiling as the JSON case |
| prefetch gate → `if False:` | parsed and packed sources are pre-read anyway; the budget stops being valid |
| byte budget `>` → `>=` | a file that exactly fills the remaining budget is skipped |

That last one is a **pre-existing off-by-one blind spot**: the neighbouring cases cover comfortably-inside and far-too-large, so nothing observed equality.

Two mutations are deliberately *not* claimed. Deleting `reader.chunk = True` is equivalent — every Agno reader already defaults to chunking on. Replacing the fallback document's `uuid4` id with a constant is also equivalent: `ChromaDb.upsert` derives its vector id as `md5(f"{document.id}_{content_hash}")` and `Knowledge._build_content_hash` mixes in the per-file path, so an identical document id still yields distinct vector ids. An earlier revision of this PR asserted otherwise and that assertion was removed rather than reworded — a test implying a hazard that cannot occur is worse than no test.

## Checks

- 452 passed across the readers, manager, resumable-refresh, git-source, chunking and import-graph suites.
- `uv run pre-commit run --all-files` clean; `uv run tach check --dependencies --interfaces` clean.
- No production diff, so no Tach boundary or module-table change is needed.


basnijholt:
Addressed the one finding (Qodo skill insight, `knowledge/readers.py` undocumented), partly — and the split is deliberate.

**Added the `CLAUDE.md` row** (63c207ff8). "Where does a knowledge file get split into chunks" is genuinely hard to locate from the tree, which is what that table is for, and it matches what the two sibling extractions immediately before this one did: #1715 added the `knowledge/collections.py` row, #1716 added `knowledge/git_source.py`, each in the extraction PR itself.

**Did not add a `docs/architecture/index.md` row.** The finding's evidence line — "the existing module tables in `docs/architecture/index.md` and `CLAUDE.md` do not include it (they list other modules but not `knowledge/readers.py`), violating the requirement that architecture docs list all source modules accurately" — does not hold for that file. Its *Key Internal Modules* table lists **zero** of the 21 modules under `src/mindroom/knowledge/`: not `manager.py`, not `registry.py`, not `refresh_runner.py`. It is scoped to the inbound-turn pipeline (`turn_controller.py`, `coalescing.py`, `delivery_gateway.py`, …), and its *Data Flow* section below the table walks that same pipeline. A knowledge-indexing row would be the first of its kind there, so the accurate fix for that file is not a one-line append.

For the record on the underlying rule: `CLAUDE.md` lists 5 of 21 knowledge modules, so neither table is an exhaustive index of source modules, and the review's own relevance rating (`● Weak`) plus its note that the same request was rejected in #1718 both point the same way.

basnijholt:
Addressed all three findings in `fd53f292a`. The central one was correct and my justification was wrong; I verified the mechanism myself rather than taking it on report.

**1. The rationale was false — withdrawn, and the code changed with it.** Both paths take their reader from one `build_reader` call: `manager.py:732` (insert) and `manager.py:853` (prefetch), identical arguments. For a type `build_reader` leaves alone it returns the shared factory instance unmodified, so both paths read through the *same* unconfigured reader and chunk identically. Widening one gate without the other costs at most a missed optimization, never a wasted embed. The drift story is impossible, and it was asserted in the two docstrings, the constant's comment, and the PR body. All four are corrected; the PR body now records the claim as withdrawn rather than quietly dropping it.

**2. Separated, per your leaning, and I agree with your reason over mine.** Two expressions with equal values today but different meanings must not be unified — merging them is the false-DRY error, and my finding-1 story was the only thing that made them look like one question. Concretely, welding them created the hazard you describe: adding `PDFReader` to the shared tuple to fix oversized embeds would silently also enable PDF prefetch, parsing every PDF twice per refresh and handing `max_chunk_text_bytes` — documented against a size on disk — an archive whose extracted text dwarfs it.

The gates are now independent. `build_reader` keeps its own inline check for "does MindRoom's text chunking apply". The prefetch predicate is renamed `reader_uses_configured_chunking` → **`reader_decodes_plain_text`**, named for the single property that actually underwrites both of its preconditions: a reader that decodes text rather than parsing a document or unpacking an archive is cheap to re-read *and* has its decoded text bounded by its size on disk. Both sites now state their own reason and say explicitly not to merge them again. `_chunk_texts_for_prefetch`'s docstring previously said "only the text-like readers MindRoom configures chunking for are pre-read", which was the same conflation in the pre-existing code; it now gives the two real reasons and notes that chunk equality with the insert path holds by construction, not by agreement between gates.

**3. Coverage added — your mutation table reproduced, then killed.** `tests/test_knowledge_readers.py`, 26 cases. Each mutation, run against that file:

| mutation | before (full manager suite) | now |
|---|---|---|
| predicate → always `True` | 0 failures | **fails** |
| predicate → always `False` | 0 failures | **fails** |
| chunking gate → `(TextReader,)` | 0 failures | **fails** |
| `chunking_strategy` never assigned | — | **fails** |

It pins that text-like readers carry *this base's* strategy object at the configured size across `.md .markdown .txt .py .yaml .html`/unknown/no-suffix; that `.csv .xlsx .docx` keep their factory chunking untouched; that `.json` keeps structured `FixedSizeChunking` under the decode-tagging subclass; that the malformed-JSON fallback reader carries base chunking; and both directions of `reader_decodes_plain_text`. Your framing was the argument that landed it: a 150-line bespoke probe to establish equivalence, and nothing left behind to hold it.

**4. `CLAUDE.md` row** — already present in `63c207ff8`, pushed before the review ran against `d700ac5ea`. No action.

Full targeted run green (413 passed across the readers, manager, resumable-refresh, chunking, and import-graph suites), `pre-commit run --all-files` clean, `tach check` clean. No force-push; `fd53f292a` is a follow-up commit and #1728 is now in the base.

basnijholt:
Both findings addressed in `c47b64aa0`. I reproduced the five survivors before changing anything — all five did survive, exactly as reported.

**1. The tests asserted the wrong thing, and the cause was design, not missing cases.** They injected an already-built strategy and asserted `reader.chunking_strategy is chunking`, which proves only that the object handed in came back out. So `chunking_strategy_for_base` was bypassed entirely, in-place mutation of the strategy was invisible through the identity check, and no field beyond `chunk_size` was checked at all.

They now start from an authored `Config` (so config → strategy is covered rather than skipped), assert every reader field that can move a boundary **by value** — strategy type, `chunk_size`, `overlap`, `min_chunk_fill_ratio`, `chunk`, `encoding` — and read real text so the boundaries are observed instead of inferred: chunk lengths pin `chunk_size`, and the characters each chunk shares with the next pin `overlap`.

**2. The rationale was overclaiming for the second time, so it is now a check rather than an inference.** You are right that the bound depends on the encoding and that type membership cannot settle it: `TextReader(encoding="utf-16")` is genuinely cheap to re-read and still decodes 202 bytes of CJK into 300 UTF-8 bytes, under-budgeting prefetch. Latin-1 does the same at 1 byte → 2.

Rather than write a third inferred rationale, `reader_decodes_plain_text` → **`reader_rereads_within_file_size`**, which admits a reader only when its type is decode-only **and** its encoding cannot expand when decoded. Behavior today is unchanged: `None` is how both readers spell UTF-8 (`self.encoding or "utf-8"` in `text_reader.py:45` and `markdown_reader.py:74`), and every reader the factory returns has `encoding=None`. An unknown codec is refused rather than assumed. The docstring now describes the check it performs.

**Mutation results.** Every mutation run against `tests/test_knowledge_readers.py` (32 cases):

| mutation | before | now |
|---|---|---|
| `chunking_strategy_for_base` → `1/0` | survived | **killed** |
| `_configure_text_reader` zeroes supplied overlap | survived | **killed** |
| `_configure_text_reader` sets `encoding="utf-16"` | survived | **killed** |
| `text_fallback_reader` → `chunk=False` | survived | **killed** |
| `_FallbackAwareJSONReader` → plain `JSONReader` | survived | **killed** |
| predicate → always `True` | killed | killed |
| predicate → always `False` | killed | killed |
| chunking gate → `(TextReader,)` | killed | killed |
| `chunking_strategy` never assigned | killed | killed |
| encoding check dropped | — | **killed** |
| `latin-1`/`utf-16` admitted as size-preserving | — | **killed** |
| unknown codec treated as safe | — | **killed** |
| `reader.chunk_size` never set | — | **killed** |
| `reader.chunk = True` removed | — | **killed** |

The last one was initially an *equivalent* mutant, not a gap: every Agno reader defaults to `chunk=True`, so deleting the assignment changes nothing observable through the factory. It is worth distinguishing anyway — `chunk=False` is the largest boundary change possible, one vector per file — so one test monkeypatches the factory to hand over a `TextReader(chunk=False)` and asserts `build_reader` forces it back on. That pins the defensive intent of the line rather than today's default.

Equivalence re-verified after the predicate change, since `build_reader`'s control flow moved again: constructed readers across 13 suffixes × 3 chunk configs plus the fallback readers are identical to pre-PR, with the two moved classes' `__module__` the only difference.

420 passed across the readers, manager, resumable-refresh, chunking, and import-graph suites; `pre-commit run --all-files` and `tach check` clean. No force-push.

basnijholt:
Fixed in `ffcd0175c`. Both mutations reproduced first — they did survive 43/43 — and the pattern diagnosis turned out to be the more valuable half of this finding.

**The codec cases now assert the property, not the spelling.** Each of `utf-8`, `UTF-8`, `utf8`, `U8`, `utf_8`, `cp65001` is asserted admitted **and** its decode measured against the source file's size on disk; `utf-8-sig` is admitted and shown to shrink by exactly `len(codecs.BOM_UTF8)`; and the UTF-16 case asserts both the refusal *and* that the decode really would have outgrown the file, so the refusal is demonstrated warranted rather than merely conservative. A 100-character CJK sample carries it: three UTF-8 bytes per character against two in UTF-16, which is the inequality the check exists for.

**Then I asked your question of every other assertion in the file, and found three more one-directional properties.** All three survived their mutations before the fix:

| survivor found by the audit | what it would have done |
|---|---|
| `chunking_strategy_for_base` returns a hardcoded `137/11` | config ignored entirely; passes because every case used one chunk size |
| `reader.chunk_size = 137` hardcoded | same blind spot one layer down |
| `_FallbackAwareJSONReader` tags *every* read as a parse failure | diverts the entire corpus to the text fallback |

Fixes: the strategy is checked against two authored configs; a second base at 512/64 asserts both the reader's configuration and the boundaries a real read produces (`[512, 512, 304]`, with 64 shared characters per seam); and valid JSON is asserted to still read as structured documents.

**Full mutation table — 19 tried, 19 fail.** Round-3 pair (empty encoding set, raw-spelling membership), my three audit finds, and all fourteen from earlier rounds re-verified as regressions, including overlap zeroing, forced `utf-16`, fallback `chunk=False`, plain `JSONReader`, dropped encoding check, unknown-codec-treated-as-safe, and `min_chunk_fill_ratio=0.25`. I re-ran the mutations a second time after the formatter hooks reworked two lines, since a reformat can silently move an assertion.

43 cases. `pre-commit run --all-files` and `tach check` clean. No force-push.

For the record, since it bears on how much weight this file carries: three review rounds have now each found the same defect shape in it — assertions mirroring an assumed input rather than the input space — and in the third round my own sweep found more instances than the review did. The habit, not the individual cases, is the fix.

basnijholt:
All six addressed in `966cce70d`, plus my answer on 7. I reproduced every mutation first — all seven survived — and for finding 1 I verified the *harm*, not just the survival: `ReaderFactory.get_reader_for_extension(".md") is ...(".md")` returns `True`, and `ChunkingStrategy._generate_chunk_id` returns `f"{document.id}_{chunk_number}"`, so both consequences you describe are real.

**1. `deepcopy` — blocking, fixed.** Two bases at 137/11 and 2000/0 now assert distinct readers, distinct strategies, their own sizes, and that the factory's cached instance is exactly as it was found. Your concurrency point is what makes this more than hygiene: the source-root lock is per `KnowledgeSourceRoot` and prefetch reads run under `asyncio.to_thread`, so it is a live data race on a shared object rather than a sequential ordering quirk. The JSON branch gets the same copy-before-use assertion one level down, where aliasing is inert today.

**2. Prefetch gate — blocking, fixed.** You are right that the predicate had tests and its *application* did not. A `.csv` beside a `.md` in `test_knowledge_resumable_refresh.py` now exercises both answers where the gate actually runs, so `if False:` dies. I also swept the neighbouring prefetch tests as you asked; the byte-budget cases already pin both directions (oversized file skipped **and** smaller files behind it still read), so the gate was the only one-directional property there.

**3. Line/column — fixed.** The sample's broken key is indented to column 6, so `(3, 6)` can now fail if the two are swapped.

**4. Fallback document id — fixed.** Two malformed sources are asserted to produce disjoint chunk ids, and the caller's `name` is asserted to survive, so both that mutation and `name=str(file)` die.

**5. Docstring — fixed.** "this same `build_reader` call" → an identical call to the same function producing separate objects with the same configuration. The topic sentence now says "size-preserving text" rather than just "text".

**6. PR body — rewritten** to the shipped design, including the withdrawn duplication claim, the enforcement, and the 25-mutation table. (Verified the edit landed; `gh pr edit --body-file` silently no-opped on me earlier in this PR.)

**7. My answer: keep the enforcement, label it, and cut the prose — which is what I did.** Your unreachability finding is correct and I re-verified it independently: no module outside `readers.py` names an Agno reader class, no `encoding=` reaches a reader anywhere in `src/`, no config knob, plugins cannot supply readers.

I disagree with the cost framing, though, and that is what decides it. Of the ~25 lines, about 10 are logic (a frozenset and a six-line helper) and the rest were docstring. So the expensive part was prose — and prose is precisely what over-claimed in rounds 1, 2 and 3. I cut that docstring to a third of its length and made it state plainly that the encoding branch is reachable only from code that does not exist yet.

I rejected both alternatives for reasons, not by default. Deleting the check and leaving a comment relies on a future author reading `max_chunk_text_bytes`'s precondition — in a different module — at exactly the moment they add an encoding option; the failure if they don't is silent memory-bound violation. Narrowing to `encoding is None` is tempting since it drops `codecs` and six test cases, but it would *build in* the silent throughput loss that the empty-`frozenset` mutation was flagged for one round earlier: an explicitly-set `utf-8` would be refused and prefetch would quietly switch off. Paying ten lines to avoid a silent failure in both directions is the right trade.

**25 mutations, 25 fail.** The six from this round, plus all nineteen earlier ones re-verified as regressions. 46 cases; 478 passed across the readers, manager, resumable-refresh, git-source, chunking and import-graph suites; `pre-commit` and `tach check` clean. No force-push.

Fourth round, fourth instance of one shape — assertions pinning one direction of a property. This time it was displaced a module over, into the call site rather than the predicate. Noting it plainly: the lesson isn't the individual cases, it's that "which direction does this observe?" belongs on every assertion I write, including the ones in files I'm only touching in passing.

basnijholt:
Fixed in `4d43d3ed3`. All four reproduced first — all four survived — and I read the Agno source rather than accept the mechanism, which mattered because one finding was itself wrong.

**3 — the overwrite claim is false, and the test built on it is gone rather than reworded.** Verified directly in `agno/vectordb/chroma/chromadb.py`: `base_id = document.id or md5(...)`, then `doc_id = md5(f"{base_id}_{content_hash}")`, and `Knowledge._build_content_hash` mixes in the per-file path. So two malformed files with an identical `document.id` still get distinct vector ids. There is no overwrite, at any level — across files the content hash differs, within a file the chunk number differs, and `ChromaDb.upsert` de-duplicates identical ids in a batch anyway.

I could not find any observable consequence of a constant fallback id, so I did not restate the rationale in weaker terms: the assertion is deleted and the mutation is now listed as **equivalent**, with the reason, in both the test docstring and the PR body. A test implying a hazard that cannot occur is worse than no test — it would have outlived both of us as evidence for something false.

What replaced it is a property that *is* observable, which I checked before asserting it: `ChromaDb.upsert` stores `document.name` as the `name` metadata field (`metadata["name"] = document.name`) and there is a `collection.get(where={"name": {"$eq": name}})` lookup. So a read that substitutes its own name makes a file's chunks findable under the wrong identifier. Asserted for the fallback path and the valid-JSON path.

**1 — JSON chunking was only observed below the chunk size.** A 6,013-character element now splits into 3 documents, none above 5,000, with `reader.chunk is True` asserted.

**2 — `name=name` was deletable because the valid-JSON case never passed a name.** It now passes `"valid.json"` and asserts every returned document keeps it; without the argument they come back as the stem `valid`.

**4 — in scope, and it corrects my own report.** I said last round that the neighbouring budget cases pinned both directions. They pin admit and refuse; **neither observes equality**, which is exactly the gap. Two 2,000-byte files under a 4,000-byte budget now assert both are read, so `>` → `>=` fails. One line of test for a real off-by-one, pre-existing or not.

**28 mutations tried: 27 fail, 1 equivalent by design.** All 19 earlier ones re-verified as regressions. 47 cases; 480 passed across the readers, manager, resumable-refresh, git-source, chunking and import-graph suites; `pre-commit` and `tach check` clean. PR body updated and verified against the live body.

**On the diagnosis — the companion question is the right one and it is sharper than mine.** "Which direction does this observe?" would not have caught any of these three: each had both directions present and used a single *value* on one side of a boundary. JSON chunking below 5,000, a name never passed, a budget never equal. That is the same root as the `137/11`-only blind spot, which I found and then failed to generalise past the one function it appeared in. Both questions now, on every assertion: which direction, and which value.

basnijholt:
Rescoped: **the extraction is reverted, the tests stay.** Head `63d694128`, retitled, body rewritten. `git diff origin/main -- src/ tach.toml CLAUDE.md` is empty.

I did not argue against this. The accounting is fair, and the claim I had been leaning on — that these tests needed the new module — was simply false: `KnowledgeManager` already has the behavior, so the coverage was always obtainable. Once that fell, the only thing left arguing for the merge was the work already spent.

**Route taken: the private seam, and the end-to-end route was ruled out by measurement rather than by convenience.** Removing the `deepcopy` from `_build_reader` leaves *both* existing end-to-end suites green — `test_knowledge_resumable_refresh.py` and `test_knowledge_manager.py`, verified. The reason is structural: within one refresh, every `_build_reader` call re-applies that manager's own chunking before the reader is used, so a sequential refresh is always internally consistent. Corruption needs two bases' refreshes to interleave, which is reachable in production (per-root locking, prefetch reads on worker threads) but cannot be staged deterministically. An end-to-end test therefore cannot hold this invariant; observing the constructed reader can. The boundary tests use the same seam for consistency, and the end-to-end boundary assertion already on main is not duplicated.

**17 mutations of unmodified `manager.py`, 17 killed** — the `deepcopy` pair, chunking-strategy construction and assignment, `chunk_size`, in-place overlap zeroing, `min_chunk_fill_ratio`, the text gate, JSON tagging and its chunking and `name=`, the fallback's chunk flag and `name=`, the prefetch gate, and the byte budget's exact-fit comparison. Full table in the body.

Two mutations are explicitly **not** claimed, both equivalent: deleting `reader.chunk = True` (every Agno reader defaults to chunking on) and a constant fallback document id (`ChromaDb.upsert` hashes the document id together with a per-file content hash, so identical ids still yield distinct vector ids). An earlier revision asserted the second was a real overwrite hazard; that assertion is deleted rather than reworded.

Dropped rather than ported: the six codec cases, which guarded the extracted module's own unreachable enforcement; the chunk-flag case, as equivalent; and the direct chunking-strategy case, whose property the two-config boundary tests cover.

Kept as its own file rather than folded into `test_knowledge_manager.py`, which is already ~8,000 lines.

452 passed across the six knowledge suites; `pre-commit` and `tach check` clean.

PR #1792, opened 2026-08-05
Let compaction summaries condense stale context

basnijholt:
## Summary

- retain prior-summary context needed to continue the work while allowing obsolete detail to fall out
- keep retained wording stable when practical, but condense it when needed
- incorporate relevant new information and preserve exact technical facts while they remain load-bearing
- keep existing smaller-input retry and output-truncation safeguards unchanged

## Verification

- `uv run pytest -q -x -n auto --no-cov`
- `uv run pytest tests/test_compaction_invariants.py tests/test_history_summary_call.py tests/test_history_compaction_rewrite.py tests/test_vertex_claude_context_guard.py -q -x -n auto --no-cov`
- `uv run pre-commit run --files src/mindroom/prompts.py`

### DIFF
--- a/src/mindroom/prompts.py
+++ b/src/mindroom/prompts.py
@@ -337,9 +337,10 @@ Your job is to produce one merged handoff summary as plain text.
 Return only the summary text.
 
 Rules:
-- Retain the information from <previous_summary> needed to continue the current work, but condense it rather than copying wording verbatim.
+- Retain the information from <previous_summary> needed to continue the current work.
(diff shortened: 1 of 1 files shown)


basnijholt:
Addressed both valid old-head review points in a052102d6: prior-summary-only needed context remains eligible for retention, and retained wording stays stable when practical for prompt-cache reuse. I did not add a mocked model-behavior test because it cannot prove prompt compliance; the focused compaction suite and hooks remain green.

PR #1795, opened 2026-08-06
Matrix event journal, projection, and outbox (foundation; cutover pending)

basnijholt:
Implements the prototype stage of `docs/superpowers/plans/2026-08-05-matrix-event-journal-projection.md`, up to and including its feasibility decision. **Draft: the production cutover has not happened yet, so the old owners are all still active.** See "What is not done" below.

Depends on mindroom-ai/mindroom-nio#49. `pyproject.toml` already pins `mindroom-nio[e2e]==0.37`, so CI cannot resolve dependencies until that release exists.

## What this adds

A durable store with one owner per fact, on SQLite and PostgreSQL behind the same SQL:

- **`event_journal/`** — the journal (what was accepted and still owes work), the visible-message projection (what a conversation currently looks like), and the response outbox (what we intend to send and whether it landed). Runtime code only ever receives a principal-bound view, so operational methods take no `principal_id` and reaching another bot's rows is not expressible.
- **`matrix/journal_ingress.py`** — durable admission on nio's admission callback. Commits before telling nio the event was accepted; a failure raises `CallbackNotAcceptedError` so the event is redelivered rather than lost.
- **`pending_event_worker.py`** — drains pending events in receipt order per room, concurrently across rooms. No durable `running` state, so a crash mid-turn leaves the event eligible again instead of stranded.
- **`matrix/conversation_hydration.py`** — one-time hydration per conversation per membership, and the point refetch owed after the visible revision of a message is redacted. Same code path for both, so there is one implementation of "ask the server what this looks like now".
- **`matrix/conversation_reads.py`** — the strict/non-strict split. Strict waits and fails closed; non-strict never waits. Neither can see a redacted revision, because the store never returns one.
- **`response_delivery.py`** — enqueue, claim, send, acknowledge. The claim commits before network I/O, which is what stops a restarted turn from generating different content and sending it under a transaction ID the homeserver already accepted.

## A design error the live server caught

The plan required every recursive relations page to report `recursion_depth >= 3`. Against a live Tuwunel that rejects ordinary conversations.

The two homeservers mean different things by the field. Synapse returns the constant `3` — the depth it is *willing* to traverse. Tuwunel returns the depth of the deepest event it *actually returned*, so a root with one threaded reply and one edit of that reply reports `1`. Any floor above zero rejects complete pages on one server while proving nothing on the other.

MindRoom now requires only that a non-empty page reports the field at all. That still catches the failure worth catching — a server that ignores `recurse` and silently returns direct children omits it — and the plan has been corrected.

## Tests

211 new tests, every store-level one running on **both** SQLite and PostgreSQL.

- **Store contract** — principal isolation, deduplication, edit reduction (including pre-original cross-sender edits, timestamp ties, and the sender-keyed unresolved-edit rule that stops a stranger evicting the author's edit), redaction, bounded cursor paging, membership epochs, outbox.
- **Crash matrix** — all nine boundaries from the plan. Each produces one terminal turn and at most one visible response, and boundaries after the model result is durable never re-run the model.
- **Live proof** (`tests/manual/event_journal_live_proof.py`) — spins up a disposable Tuwunel and checks what CI cannot see: relation traversal, redacting the visible edit, 25-edit churn, deterministic transaction reuse, device-scoped deduplication, and bounded history exhaustion. All green.

The PostgreSQL fixture deliberately uses a **glibc** image. The existing Alpine fixture cannot fail on a missing `COLLATE "C"`, because every musl locale sorts like C. Removing the pin makes the two ordering tests fail, which was verified rather than assumed.

## Measured feasibility decision

Reproducible via `tests/manual/event_journal_measurements.py`:

| Measurement | Result | Target |
| --- | --- | --- |
| Durable admission, p95 | 0.14 ms | under 50 ms |
| Bounded conversation read, p95 | 0.19 ms | under 50 ms |
| Writer-queue wait, p95 | 5.8 ms | under 100 ms |
| SQLite lock failures, 50 concurrent conversations | 0 | zero |
| Conversation read query plan | covering index | indexed |
| Replacement / replaced source | 3,202 / 17,877 lines | net negative |
| Projected net change | −14,675 lines | materially net negative |

No compatibility facade, second writer, retained edit chain, or second recovery classifier was needed, so no stop condition was triggered.

## What is not done

The production cutover. `bot.py` still uses `dispatch_obligations/`, `matrix/cache/`, and the conversation cache; nothing in `src/` calls the new owners yet, which is why they appear in `vulture_whitelist.py`.

The remaining risk is not in these primitives. It is in the ingress restructuring: coalescing, deferred turn settlement, and streaming currently run as background work that the pending worker would have to own and await, and that reshapes `TurnController`'s relationship to background tasks.

The full existing suite passes unchanged, so this is purely additive today.

## Summary by Sourcery

Introduce a new principal-scoped Matrix event journal with projection, hydration, and deterministic outbox delivery, proven on SQLite and PostgreSQL but not yet wired into production flows.

New Features:
- Add principal-bound event journal store with shared SQLite and PostgreSQL backends to durably record inbound Matrix events and pending work.
- Introduce a visible-message projection layer that maintains one row per logical message with edit reduction, redaction handling, and bounded cursor-based conversation reads.
- Add Matrix conversation hydration and point-refetch logic that rebuilds threads and rooms from the homeserver once per membership epoch, including strict and non-strict read modes.
- Implement a deterministic response outbox and delivery helper that support crash-safe initial and final message deliveries with idempotent transaction IDs.
- Add a pending-event worker that drains journal-backed pending events per room in receipt order while isolating failures and surviving crashes.

Enhancements:
- Define a backend-agnostic SQL schema and identity helpers so SQLite and PostgreSQL share the same journal, projection, and outbox behavior.
- Add a PostgreSQL journal test fixture using a glibc image and locale to validate byte-order text collation and cursor ordering across backends.
- Extend test infrastructure with a parametrized journal_store fixture to run store-level tests against both SQLite and PostgreSQL.
- Whitelist new, not-yet-wired owners in the static analysis configuration to keep the existing production path unchanged during the prototype stage.

Build:
- Bump mindroom-nio dependency to version 0.37 to consume new recursive relation capabilities required by the hydrator.

Documentation:
- Add a detailed architecture and cutover plan for the Matrix event journal, projection, hydration, and outbox, including invariants, feasibility criteria, and crash matrix boundaries.

Tests:
- Add comprehensive store contract tests covering principal isolation, admission semantics, edit reduction, redaction behavior, membership epochs, bounded paging, and outbox behavior on both SQLite and PostgreSQL.
- Add crash-matrix tests that simulate nine distinct crash boundaries from admission through delivery acknowledgement to prove turn uniqueness and single visible response behavior.
- Add conversation hydration and read-mode tests that exercise relation traversal, recursion-depth handling, room history hydration, point refetch after redaction, and strict vs non-strict reads.
- Add journal ingress tests validating provenance mapping, kind classification, durable admission semantics, replay fidelity, and pending worker behavior.
- Add manual measurement and live-proof scripts to benchmark journal performance and validate homeserver assumptions such as recursion depth semantics, edit redaction behavior, transaction deduplication, and bounded history exhaustion against a live homeserver.

Chores:
- Extend pytest Postgres lifecycle management to create and clean up a separate glibc-based journal container alongside the existing cache Postgres container.


basnijholt:
Closing as fully superseded by #1800.

Every useful foundation commit from this PR is part of the #1800 implementation lineage. The only commit unique to this branch is the temporary rollback of the ingress cutover, made while the production cutover was still incomplete. #1800 subsequently completed that cutover, expanded the journal implementation and tests, and removed the old competing owners.

There is nothing left in this PR to merge or extract.

PR #1812, opened 2026-08-12
fix: stop the startup sweep from deleting live approval claims

basnijholt:
## Summary

- own an approval transaction from before its claim row is written until the send registers
- keep a wall-clock horizon as the backstop for a row whose owner is gone from memory
- keep sweeping instead of giving up, and get louder instead of quieter
- report what every sweep pass walked, skipped, and dropped

## Why

`request_approval` commits the durable approval row (`_claim_card`) before the send that owns it exists; the send only registers itself in `_active_approval_sends` inside `_send_and_bind_waiter`. `_send_is_in_flight()` reads that set, so across the claim await there is nothing in memory saying the row is spoken for. The startup sweep reads exactly what a dead process leaves behind — claimed, never attempted — and deletes it in `_identified_card`, logging `approval_startup_card_dropped_never_attempted`.

That deletion is upstream of every other guard: `_forget_card` runs before the live-waiter check, before any continuation check, and the row is gone before the branch that would have protected it is reached.

The sweep is also not confined to startup. An unfinished pass reschedules itself, so a single card that can never be settled keeps it running on a 30s timer for the life of the process. One production control plane, 19 hours into its uptime, logged 17 deletions across 9 rooms in six hours with `owed_count=2` throughout. A single pass took six different rooms' rows in nine seconds.

**What this does not claim.** The deletion is not what initially wedges a conversation. In the room studied most closely, `bridge_entry` was at 18:55:19.720 and the row was still unattempted when the first sweep pass reached it at 18:59:10 — the caller had already been parked for 3m51s, and the first pass after that restart could not have run any earlier. The sweep destroys the durable card of a call that is already stuck, and takes other rooms' healthy in-flight claims with it. Why a caller parks between its claim and its send is a separate defect, still open.

## What changed

**Ownership.** `_publishing(transaction_id)` is entered before the claim write and left once the send has registered its own hold or the attempt has ended. It is a `Counter`, not a set, so overlapping scopes for one transaction cannot have the inner one release the outer one's hold. `_send_is_in_flight` consults it alongside the two existing owners.

**Horizon.** `_sweep_horizon_ns` is still taken at manager construction, now as a backstop rather than the primary guard: it covers a row whose owner died without cleanup, where the timestamp is the only remaining proof no dead process wrote it. Construction rather than first sweep, because the arming gates are late — production `rooms_and_memberships` has been observed taking 350s.

Both guards are load-bearing. Disabling the ownership hold fails `test_a_live_claim_survives_a_clock_that_went_backwards`; disabling the horizon fails `test_a_claim_this_process_made_is_left_alone_once_its_owner_is_gone` with `dropped_never_attempted=1`. Verified by mutation, not by inspection.

**Retry.** The earlier attempt cap is gone. It existed because each pass was a chance to delete live work, which is no longer true, and the arithmetic reverses: ten attempts is ~151s against a 350s startup envelope, an outage that outlasts any fixed budget is when cleanup is owed most, and the repeating sweep has been the only thing collecting rows orphaned later in a process's life. Past `_STARTUP_CLEANUP_ATTEMPTS_BEFORE_ESCALATION` the incomplete log moves from warning to error.

**Observability.** `ApprovalStartupSweep` now carries `scanned`, `skipped_after_start`, `skipped_in_flight`, `dropped_never_attempted` (excluded from equality), and one unconditional `approval_startup_sweep_finished` reports them. The healthy outcome of these guards is a pass that settles nothing, which was previously indistinguishable from a pass that never ran — and the deletion path incremented no counter at all, so those 17 production deletions appeared in no sweep total. `tool_approval_startup_discard_failed` gains `exc_info=True`; both failure lines gain `attempt`.

## Tests

`uv run pytest -q -nauto tests/test_tool_approval.py tests/test_orchestrator_runtime.py tests/test_bot_reactions_approvals.py tests/test_tool_hooks.py` — all pass. `ruff check`, `ruff format --check`, `ty check` clean on the touched files.

New:
- `test_startup_sweep_landing_inside_a_live_claim_leaves_it_alone` — drives the sweep from inside `claim_approval_card`, after the row commits and before the call returns. Verified red before the fix: the claim is deleted, the card never becomes answerable, the decision wait times out.
- `test_a_live_claim_survives_a_clock_that_went_backwards` — horizon pushed an hour into the future, so only ownership can save the claim.
- `test_a_claim_this_process_made_is_left_alone_once_its_owner_is_gone` — post-horizon row with no holder at all.
- `test_a_startup_discard_that_keeps_coming_up_short_gets_louder` — asserts the escalation actually fires, via `structlog.testing.capture_logs`.

`FakeApprovalCards` now timestamps claims with a clock like the real table (`event_journal/approvals.py:196`) instead of a counter. A counter puts every row before any process start, so the double could not fail the way the real store does. Seed helpers backdate explicitly, since they stand for a previous process's work.

## Relationship to other PRs

**#1811 should be closed or re-pointed.** Its target branch is unreachable after this: a live waiter only exists for a card this process sent, so such rows are always owned or post-horizon and the sweep skips them before `_discard_live_card` is reached. Its two new tests construct the manager and then call `request_approval`, expect `(discarded=1, failed=0)`, and get `(0, 0)` with the decision wait timing out. Merging it first is also a user-visible regression: under a backward clock step it force-expires healthy approvals, showing "Bot restarted before approval" on a bot that did not restart.

**#1807 wants this first.** Its `continuation_id` guard sits after `_identified_card`, so without this PR a detached card claimed in the same window is deleted before that guard is read. Its `create_detached_approval` awaits `_mark_attempted_then_send` directly and never builds an `_ActiveApprovalSend`, so the publishing hold added here is the only thing covering it.

Suggested order: this PR, then #1807, closing #1811.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented startup recovery from interfering with approvals currently being published.
  * Preserved approvals when requests end unexpectedly after claiming them.
  * Improved recovery behavior during clock shifts.
  * Prevented duplicate approval sends while a delivery is in progress.

* **Monitoring**
  * Added detailed startup-sweep counts for scanned, skipped, and dropped approvals.
  * Repeated incomplete sweeps now escalate from warnings to errors after ten attempts.
  * Completion logs include outstanding approval counts and sweep statistics.

### DIFF
--- a/src/mindroom/approval_manager.py
+++ b/src/mindroom/approval_manager.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import asyncio
 import json
 import threading
-import time
 from collections import Counter, OrderedDict
 from collections.abc import Awaitable, Callable, Iterator
 from concurrent.futures import Future, InvalidStateError
@@ -277,7 +276,6 @@ class ApprovalStartupSweep:
     # because a caller comparing two sweeps is asking whether the same work
     # was settled, not whether the same rows happened to be on disk.
     scanned: int = field(default=0, compare=False)
-    skipped_after_start: int = field(default=0, compare=False)
     skipped_in_flight: int = field(default=0, compare=False)
     dropped_never_attempted: int = field(default=0, compare=False)
 
@@ -297,7 +295,6 @@ class _SweepTally:
     """
 
     scanned: int = 0
-    skipped_after_start: int = 0
     skipped_in_flight: int = 0
     dropped_never_attempted: int = 0
 
@@ -408,13 +405,6 @@ class _ApprovalManager:
         # over. Counted rather than flagged so overlapping scopes for one
         # transaction cannot have the inner one release the outer one's hold.
         self._claiming_transaction_ids: Counter[str] = Counter()
-        # Rows claimed from here on belong to this process, whether or not the
-        # send that owns one has registered itself yet. Taken at construction
-        # rather than at the sweep, because the sweep is armed by startup gates
-        # that can be minutes late, and every claim made while waiting for them
-        # would otherwise look like one a dead process left behind. Backstop
-        # for the ownership above, which is exact but cannot outlive memory.
-        self._sweep_horizon_ns = time.time_ns()
         # Recovery that outlived the request that started it, held so it is
         # not garbage collected mid-flight.
         self._detached_card_writes: set[_DetachedCardWrite] = set()
@@ -577,7 +567,6 @@ class _ApprovalManager:
             discarded=discarded,
             failed=failed,
             scanned=tally.scanned,
-            skipped_after_start=tally.skipped_after_start,
             skipped_in_flight=tally.skipped_in_flight,
             dropped_never_attempted=tally.dropped_never_attempted,
         )
@@ -619,26 +608,6 @@ class _ApprovalManager:
                 transaction_id=claimed.transaction_id,
             )
             tally.skipped_in_flight += 1
-            return None
-        if claimed.created_at_ns >= self._sweep_horizon_ns:
-            # Claimed after this process started, so it is this process's to
-            # finish. The check above cannot always see it: a claim is
-            # committed before the send that owns it registers, and a sweep
-            # landing in that window would read a live claim as an abandoned
-            # one and delete the row out from under the caller still building
-            # its card. The ownership registered around the claim closes that
-            # window exactly; this closes it again for a row whose owner is
-            # gone from memory but whose timestamp still says it was never
-            # this sweep's to take. Nothing is owed either way -- the owner
-            # settles it or fails on its own path.
-            logger.debug(
-                "approval_startup_card_skipped_claimed_after_start",
-                room_id=room_id,
-                transaction_id=claimed.transaction_id,
-                created_at_ns=claimed.created_at_ns,
-                sweep_horizon_ns=self._sweep_horizon_ns,
-            )
-            tally.skipped_after_start += 1
             return None
         identified = await self._identified_card(room_id, claimed, tally=tally)
         if identified.card is None:
--- a/src/mindroom/approval_manager.py
+++ b/src/mindroom/approval_manager.py
@@ -277,7 +277,6 @@ class ApprovalStartupSweep:
     # was settled, not whether the same rows happened to be on disk.
     scanned: int = field(default=0, compare=False)
     skipped_in_flight: int = field(default=0, compare=False)
-    skipped_live_waiter: int = field(default=0, compare=False)
     dropped_never_attempted: int = field(default=0, compare=False)
 
     @property
@@ -297,7 +296,6 @@ class _SweepTally:
 
     scanned: int = 0
     skipped_in_flight: int = 0
-    skipped_live_waiter: int = 0
     dropped_never_attempted: int = 0
 
 
@@ -570,7 +568,6 @@ class _ApprovalManager:
             failed=failed,
             scanned=tally.scanned,
             skipped_in_flight=tally.skipped_in_flight,
-            skipped_live_waiter=tally.skipped_live_waiter,
             dropped_never_attempted=tally.dropped_never_attempted,
         )
 
@@ -634,25 +631,13 @@ class _ApprovalManager:
             return None
         if identified.card.resolution is not None:
             return await self._redeliver_recorded_resolution(pending, identified.card)
-        if self._live_waiter_for_card(pending.card_event_id) is not None:
-            # This process sent this card and is still waiting on the answer.
-            # A waiter only ever gets here one way -- ``_bind_live_waiter``,
-            # reached from a fresh send -- so a card with one is not a card a
-            # dead process left, it is live work of this one's. Expiring it
-            # tells a user their request was cancelled by a restart that did
-            # not happen, seconds after they were asked for a decision.
-            #
-            # Nothing is owed either: the waiter settles the card on a click,
-            # a denial, or its own timeout. Counting it would have every
-            # pending approval hold the sweep open until the human answers.
-            logger.debug(
-                "approval_startup_card_skipped_live_waiter",
-                room_id=room_id,
-                transaction_id=identified.card.transaction_id,
-                card_event_id=identified.card.card_event_id,
+        live_waiter = self._live_waiter_for_card(pending.card_event_id)
+        if live_waiter is not None:
+            return await self._discard_live_card(
+                waiter=live_waiter,
+                reason=_STARTUP_DISCARD_REASON,
+                resolved_by=transport_sender,
             )
-            tally.skipped_live_waiter += 1
-            return None
         result = await self._discard_matrix_only_card(
             pending=pending,
             transaction_id=identified.card.transaction_id,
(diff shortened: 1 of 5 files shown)


basnijholt:
All three addressed, plus the change the merge of #1811 forces. Review was against `153c2a80`; head is now `abf38f402`, rebuilt on current `main` (`3f2fe8b3`, which contains #1811).

**P1 horizon shadows #1811's live-waiter cleanup — fixed by deleting the horizon.**
Correct call, and it went further than narrowing the guard. Classifying a row by when it was written was a proxy for asking who owns it, and the exact answer is available: a request now holds its transaction from before the claim write until the send registers its own hold or the attempt ends (`_publishing()`, a refcount so overlapping scopes cannot release each other), and `_send_is_in_flight` consults it. `main` is merged in and `test_startup_sweep_rejects_live_approval_and_releases_waiter` passes alongside the new tests — no timeout, no dead code.

**P1 wall-clock horizon is unsafe — gone entirely.**
No timestamp classification remains; `_sweep_horizon_ns` and the `time` import are deleted. The backward-step regression stays, with the clock moved into the card store: `FakeApprovalCards.stamp_next_claim_ns()` dates the row under test an hour before every other row, and it survives on ownership alone (`test_a_live_claim_is_protected_by_ownership_not_by_its_timestamp`).

Worth stating what the horizon additionally covered and why losing it is fine: a row this process claimed and then abandoned. Deleting that row is not a bug — it is unattempted, so nothing reached the room, and collecting it is the sweep's job.

**P2 warning untested — the behaviour it tested is gone; the replacement is asserted.**
`approval_startup_sweep_abandoned` no longer exists. The attempt cap was removed: it was justified only while each pass could delete live work, and once that is false the arithmetic reverses. Ten attempts is ~151s against a startup envelope observed at 350s, an outage outlasting any fixed budget is when cleanup is owed most, and the repeating sweep was the only thing collecting rows orphaned later in a process's life. The sweep keeps retrying and gets louder instead: past `_STARTUP_CLEANUP_ATTEMPTS_BEFORE_ESCALATION` the incomplete line moves warning to error. `test_a_startup_discard_that_keeps_coming_up_short_gets_louder` asserts the levels via `structlog.testing.capture_logs`.

Also added since the reviewed commit: one unconditional `approval_startup_sweep_finished` carrying `scanned / discarded / owed_count / skipped_in_flight / dropped_never_attempted`. The healthy outcome of this guard is a pass that settles nothing, previously indistinguishable from a pass that never ran, and the deletion path incremented no counter at all.

Both guards were checked by mutation rather than inspection: disabling the ownership hold fails the two live-claim tests with a decision-wait timeout.

`pytest -nauto` over `test_tool_approval.py`, `test_orchestrator_runtime.py`, `test_bot_reactions_approvals.py`, `test_tool_hooks.py` — exit 0. `ruff check`, `ruff format`, `ty check` clean.

One correction to the PR's original framing, unrelated to these three: the deletion is not what initially wedges a conversation. In the production room studied most closely, `bridge_entry` was at 18:55:19.720 and the row was still unattempted when the first sweep pass after that restart reached it at 18:59:10 — the caller had already been parked 3m51s, and no earlier pass existed. This PR stops the sweep destroying the durable card of an already-stuck call and taking other rooms' healthy in-flight claims with it. Why a caller parks between its claim and its send is a separate open defect.

basnijholt:
Head is now `a2e198440`. Both remaining blockers were raised against `7680784d4` and are resolved by the two commits after it; the `has_live_work` gap is fixed in this push.

**Current-base incompatibility — resolved.** The branch now contains live `main`:

```
$ git merge-base --is-ancestor 3f2fe8b3b HEAD && echo YES
YES: live main 3f2fe8b3 is an ancestor of HEAD
$ uv run pytest tests/test_tool_approval.py -k rejects_live_approval_and_releases_waiter -q
1 passed
```

#1811's regression passes against the real merge, not a synthetic one. The PR body's "close #1811" line is gone — it was written before #1811 merged, and reconciling was the right call.

**Clock protection ending too early — resolved by removing the clock.** There is no timestamp guard left:

```
$ grep -rn "sweep_horizon\|created_at_ns >=\|skipped_after_start" src/mindroom/ | wc -l
0
```

`_sweep_horizon_ns`, the `created_at_ns` comparison, the `skipped_after_start` counter and the `time` import in `approval_manager.py` are all deleted. Your probe's failure mode — a bound waiter skipped by the horizon, `failed=1`, `skipped_in_flight=0`, no terminal edit — needs the horizon to exist. After a bind the row carries a `card_event_id`, so `_identified_card` returns it and #1811's live-waiter branch runs, which is what that passing regression exercises.

That also answers the wider point directly: after publish completes, ownership is not the guard and does not need to be. The row is acknowledged, `_pending_by_card_event` holds the waiter, and #1811 settles it. `_publishing` covers exactly the window where nothing else can: claim write to send registration.

What the horizon additionally covered was a row this process claimed and then abandoned. Deleting that row is correct — it is unattempted, nothing reached the room, and collecting it is the sweep's job.

**`has_live_work()` blind spot — fixed in `a2e198440`.** It counted `_pending_by_card_event`, `_resolving_card_event_ids`, `_active_approval_sends`, `_post_cancel_cleanup_tasks`, but neither `_claiming_transaction_ids` nor `_detached_card_writes`, so `initialize_approval_store` could still replace a manager mid-claim. Both now count, with `test_a_request_publishing_a_claim_counts_as_live_work` asserting it from inside the claim write.

**Accepted, not fixed.** A parked caller holds `_publishing` for as long as it is parked, so its row is permanently `skipped_in_flight` rather than deleted. Wedged rows accumulate until restart. Better than destroying them, and the row is the only artifact that could later settle the call, but it is a real consequence worth stating: this PR does not unwedge `!96Pf`, it stops the collateral damage. The parked claim itself is a separate defect.

`pytest -nauto` over `test_tool_approval.py`, `test_orchestrator_runtime.py`, `test_bot_reactions_approvals.py`, `test_tool_hooks.py`: exit 0. `ruff check`, `ruff format --check`, `ty check` clean.

basnijholt:
Confirmed and fixed in `d356046d3`. This one was real and I had it wrong.

**The branch cannot fire on a dead process's card.** `_pending_by_card_event` is written in exactly one place, `_bind_live_waiter` (`approval_manager.py:973`), reached from two call sites (`:860`, `:913`) that are both in the publication path. Nothing on the recovery/resend path binds a waiter. So a recovered card with a live waiter was sent by *this* process and is waiting on a human right now.

**Production had two of them at the moment it would have fired.**

```
container start                                     18:54:38
18:56:06.514  request_network_access  agno_tool_call_started
18:56:06.569  request_network_access  bridge_entry
18:56:46.361  request_network_access  agno_tool_call_started
18:56:46.500  request_network_access  bridge_entry
first sweep pass                                    18:59:10
```

Both inside the window, and both are the `owed_count: 2` that kept the sweep re-running all afternoon — the pre-#1811 path could not settle them either (`_claim_matrix_cleanup` returns False when the card event is in `_pending_by_card_event`, `:1974`) and counted them as owed. Merged as-is, Haonan gets "Bot restarted before approval — original request was cancelled" about two and a half minutes after being asked to decide, citing a restart that had already finished. Startup replay is when approvals get created, so the exposure window and the creation window are the same window.

**Fix.** Skip a card with a live waiter, and do not count it: the waiter settles it on a click, a denial, or its own timeout. Counting it would have every pending approval hold the sweep open until somebody answered — which is the loop this PR set out to stop. New `skipped_live_waiter` counter on the sweep summary. `_discard_live_card` is removed; nothing else reached it.

**This corrects #1811, which is on main.** Its two tests codified the expiry as intended, so they are re-pointed rather than deleted:
- `test_startup_sweep_rejects_live_approval_and_releases_waiter` → `test_startup_sweep_leaves_this_process_s_own_live_approval_alone`: asserts the sweep reports `skipped_live_waiter == 1`, the request is still pending, no Matrix edit was made, the row survives, and the human's later approval still lands.
- `test_startup_sweep_retries_live_approval_edit_that_did_not_land` → `test_startup_sweep_does_not_retry_against_a_live_approval`: a live card has no committed decision to redeliver, so repeated passes have nothing to converge on — they would only expire it on whichever attempt got its edit through.

@ the author of #1811: flagging directly, since the behaviour your tests describe is the thing being changed. Happy to split this into its own PR against main if you would rather review it separately from the sweep work.

Verified by mutation: disabling the new guard fails both re-pointed tests. `pytest -nauto` over the four files exit 0; `ruff check`, `ruff format --check`, `ty check` clean.

Issue #1796, opened 2026-08-06
Suspend response lifecycle while tool approval is pending


basnijholt:
## Problem

Tool approval is awaited inline inside the tool hook while the response runner keeps the conversation lifecycle active. An approval may remain pending for a long timeout. During that wait:

- typing and active-response state remain set
- later events for the same conversation queue behind the active response
- scheduled events may be accepted but remain undispatched
- cancellation or reload becomes the practical recovery path

Durable approval persistence is desirable. The flaw is coupling a durable pending approval to a live response coroutine and its conversation lock.

## Required boundary

- When approval is required, persist a restartable continuation containing the approval ID, exact tool call identity and arguments, conversation target, and execution identity.
- Emit the approval card, mark the turn as `waiting_for_approval`, stop typing, and release the response lifecycle lock.
- Resolve approval, denial, or expiry by enqueueing exactly one continuation through the normal per-conversation serializer.
- Execute only the exact approved call. Repeated decisions, recovery, or reload must not duplicate a side effect.
- If the requesting agent is no longer available, settle the continuation with a visible terminal result.
- Pending approvals may survive reload, but must not retain a live response coroutine or Matrix client.

Do not solve this by letting queued events bypass a still-active response lock; that would permit concurrent turns to mutate one conversation.

## Acceptance criteria

- [ ] A pending approval does not keep the conversation response lifecycle active.
- [ ] A later human event in the same conversation can dispatch while approval remains pending.
- [ ] A scheduled event targeting the same conversation can dispatch while approval remains pending.
- [ ] Approval, denial, and expiry each resume one serialized continuation.
- [ ] Duplicate approval events and recovery replay cannot execute the tool twice.
- [ ] Reload during a pending approval preserves the approval and restores or cleanly settles its continuation.
- [ ] Focused regression tests cover lock release, typing cleanup, reload, and duplicate-decision handling.

PR #1814, opened 2026-08-12
fix: key the sweep's live-approval guard on the transaction

basnijholt:
Post-merge review of #1811 + #1813 + #1812 found one defect of the same class those PRs were fixing, reachable through a door none of them closed. Fixing it exposed two more. All five fixes below are in the startup approval sweep; each reproduces a real failure and each is mutation-checked.

**This does not unwedge a stuck conversation.** Everything here governs rows and waiters that exist after an approval is claimed. The freeze happens before that point and is tracked separately — see the end.

---

## 1. A live approval could be resent, expired and deleted

`_settle_recovered_card`

#1813's guard asks whether a *card event* has a live waiter, and runs after `_identified_card` — whose job is rows **whose event id is unknown**. That state is reachable: `_acknowledge_card` is best effort by design and swallows its failure, leaving the row unacknowledged while the waiter is alive and the publishing hold already released.

Such a row reads exactly like one a dead process left mid-send, so `_card_the_frozen_transaction_recovers` re-sends the card under its own live waiter. While the homeserver still deduplicates the frozen transaction the repeat collapses onto the original; once it does not — Synapse's cache is minutes, `tool_approval.timeout_days` is 7 — it is a second card, the row rebinds to it, the guard stops matching, and the same pass expires the copy and deletes the row.

```text
[A] live waiter bound to $approval-1; row card_event_id=None
[A] after sweep: sends=2  discarded=1  rows left=[]
[A] user click -> consumed=True resolved=False   (no_stored_card)
```

Fix: ask whether *this transaction* is mine and still live, against a key that exists whether or not the acknowledge landed. `_LiveApprovalWaiter` already carries it. Checked next to `_send_is_in_flight`, before anything can resend, so the event-keyed check afterwards is redundant and goes. Moving it above the `resolution is not None` branch also stops a card mid-settlement being counted as a sweep failure.

## 2. …but skipping it stranded the row it was protecting

Leaving a live row alone is not enough when the row cannot name its event. Every decision is recorded against the event id the acknowledge was meant to store, so the user clicks Approve, nothing matches, the answer is dropped, and the card stays pending in the room. Recovery used to repair this by accident — the resend carried an acknowledge with it. Skipping the row removed the accident along with the resend.

```text
before: row event_id after sweep: [None]   click: consumed=True resolved=False   editor calls: 0
after:  row event_id after sweep: ['$approval']   click: True True   decision: approved   editor calls: 1
```

Fix: repair on purpose from what the waiter already holds — no send, no second card. A repair that itself fails leaves the row **owed** rather than done with, and logs `approval_startup_card_repair_failed`, because nothing else will come back for it.

## 3. A decision the room never saw could never be redelivered

The resolved set is what stops a card being answered twice. Adding to it on a decision that was recorded but never shown retires the only thing that would come back: the row survives, every later pass is refused by `_claim_matrix_cleanup` and counted owed, the sweep never completes again, and the card stays visibly pending until restart.

Fix: add the id only once the edit has landed. Audited all seven sites; two more were unconditional — the cancelled-waiter path ignoring `_emit_resolution`'s outcome, and shutdown ignoring whether its terminal edit landed. An edit that failed during shutdown is exactly what the next process's sweep exists to redeliver.

## 4. A store failure mid-recovery aborted the whole scan

Recovery writes back the event id it established, and that write was the only unguarded store call on the path. A store failing there raised out of `discard_pending_on_startup`: the tally went with it, and every row behind it in the page — and every room after it — was never looked at. With the sweep retrying on a timer, each pass aborted at the same row.

Fix: one `_bind_recovered_card_event` helper for both call sites, reporting owed and carrying on, as the five other store calls in this file already do.

## 5. Telemetry

`skipped_live_waiter` restored — it existed in the commit that first carried fix 1 and was lost when that fix moved from #1812 to #1813, which had no tally to put it in.

Counted the paths that were leaving no trace: `dropped_unrecoverable` (unidentifiable, absent-after-device-change) and `kept_unusable` (unusable body, no transport). Both are on `approval_startup_sweep_finished`. The no-transport path previously returned `settled=True` in complete silence, in a sweep that deletes things.

## Test hygiene

Dead scaffolding from the wall-clock horizon that #1812 removed before merge — `_backdate_to_previous_process`, `stamp_next_claim_ns`, `_seeded_rows`, the strict-monotonic claim clock — all proven inert by mutation and deleted, along with a test whose only distinguishing input was that stamp. Their comment claimed the paging cursor needs unique timestamps; it is `(created_at_ns, transaction_id)`, so it does not. Two tests now assert on the row before awaiting a send, so reverting the ownership hold fails them by assertion instead of a 5s `TimeoutError`.

New regressions, each verified to fail without its fix:

| Test | Guards |
|---|---|
| `test_startup_sweep_leaves_a_live_approval_whose_acknowledge_failed` | fix 1 — no resend under a live waiter; owed when the repair also fails |
| `test_startup_sweep_repairs_the_event_id_a_failed_acknowledge_left_unwritten` | fix 2 |
| `test_a_decision_the_room_never_saw_stays_redeliverable` | fix 3 |
| `test_a_store_failure_while_binding_owes_one_row_not_the_page` | fix 4 |

`uv run pytest -q -nauto tests/test_tool_approval.py tests/test_tool_hooks.py tests/test_orchestrator_runtime.py tests/test_bot_reactions_approvals.py` — exit 0. `ruff check`, `ruff format --check`, `ty check`, tach clean.

## Not in this PR

- **The sync-bridge fail-closed guard** was reverted at `06106a6c8`. Three blockers were raised and all three were right: the policy was evaluated twice so a stateful script could answer False then True and walk into the wedge; `running_loop_is_sync_tool_bridge()` meant "some other loop is blocked" rather than "I am its bridge"; and it shipped with no test. It returns rewritten with a `ContextVar`, single evaluation, and a test that reproduces the journal wedge, together with the `docs/configuration/index.md` note about approval-gated synchronous tools.
- **Shutdown does not drain publishing holds**, so a claim blocked across shutdown can resume and publish a stale card. Real; the fix is a shutdown recheck before the send rather than draining a hold whose holder is parked.
- **Mixed-page coverage** — one dead row and one live claim in a single pass. Nothing yet proves a skip does not abandon the rest of the page.
- **The tally invariant is unasserted**: `scanned` should equal the sum of the six outcome counters.

## The freeze

None of this is what a wedged user feels. An approval-gated synchronous tool parks before any of this code runs, because its journal write is issued from a loop whose writer task can only advance on a loop that is blocked joining it. And even when the write succeeds, the approval is awaited inline, holding typing and the conversation lock — which is [#1796](https://github.com/mindroom-ai/mindroom/issues/1796): persist a continuation, release the lifecycle, resume exactly once after the decision. That is the durable user-facing fix; this PR is the janitor.

### DIFF
--- a/src/mindroom/approval_manager.py
+++ b/src/mindroom/approval_manager.py
@@ -613,7 +613,8 @@ class _ApprovalManager:
             )
             tally.skipped_in_flight += 1
             return None
-        if self._live_waiter_for_transaction(claimed.transaction_id) is not None:
+        live_waiter = self._live_waiter_for_transaction(claimed.transaction_id)
+        if live_waiter is not None:
             # This process sent this card and is still waiting on the answer,
             # so it is live work of this one's rather than something a dead
             # process left. Keyed on the transaction, not the event: the row's
@@ -626,11 +627,20 @@ class _ApprovalManager:
             # Nothing is owed either: the waiter settles the card on a click,
             # a denial, or its own timeout. Counting it would have every
             # pending approval hold the sweep open until somebody answered.
+            if claimed.card_event_id is None:
+                # Except that a waiter cannot settle a row that does not name
+                # its event: every decision is recorded against the event id
+                # this write was meant to store, so leaving it unwritten
+                # strands the answer as surely as deleting the row would. The
+                # waiter is holding what the failed write was carrying, so
+                # finish it here rather than resend anything.
+                await self._acknowledge_card(live_waiter)
             logger.debug(
                 "approval_startup_card_skipped_live_waiter",
                 room_id=room_id,
                 transaction_id=claimed.transaction_id,
                 card_event_id=claimed.card_event_id,
+                repaired_event_id=claimed.card_event_id is None,
             )
             tally.skipped_live_waiter += 1
             return None
@@ -1033,9 +1043,15 @@ class _ApprovalManager:
                     self._cancelled_card_event_ids.discard(waiter.card_event_id)
             return
         with self._claimed_resolution(claimed_waiter.card_event_id):
-            await self._settle_waiter_with_terminal_edit(claimed_waiter, decision)
+            delivered = await self._settle_waiter_with_terminal_edit(claimed_waiter, decision)
             with self._live_lock:
-                self._resolved_card_event_ids.add(claimed_waiter.card_event_id)
+                # Only a card the room has actually been told about is finished
+                # with. Marking one resolved on a decision that was recorded but
+                # never shown retires the only thing that would come back for it:
+                # every later pass is refused by the cleanup claim, counted owed,
+                # and the card stays visibly pending until the process restarts.
+                if delivered:
+                    self._resolved_card_event_ids.add(claimed_waiter.card_event_id)
                 if mark_cancelled:
                     self._cancelled_card_event_ids.discard(claimed_waiter.card_event_id)
 
@@ -1060,9 +1076,15 @@ class _ApprovalManager:
         if claimed_waiter is None:
             return await self._wait_for_competing_terminal_decision(waiter)
         with self._claimed_resolution(claimed_waiter.card_event_id):
-            await self._settle_waiter_with_terminal_edit(claimed_waiter, decision)
+            delivered = await self._settle_waiter_with_terminal_edit(claimed_waiter, decision)
             with self._live_lock:
-                self._resolved_card_event_ids.add(claimed_waiter.card_event_id)
+                # Only a card the room has actually been told about is finished
+                # with. Marking one resolved on a decision that was recorded but
+                # never shown retires the only thing that would come back for it:
+                # every later pass is refused by the cleanup claim, counted owed,
+                # and the card stays visibly pending until the process restarts.
+                if delivered:
+                    self._resolved_card_event_ids.add(claimed_waiter.card_event_id)
             return decision
 
     async def _resolve_live_response(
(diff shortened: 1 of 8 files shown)


basnijholt:
Reverted the sync-bridge half out of this PR at `06106a6c8`. #1814 is back to the sweep fix only, which all three reviews agree is correct. The bridge guard needs rework and will come back separately rather than block this.

**The missing test is real and it was my error.** The PR body named `test_approval_is_refused_before_it_claims_on_a_sync_tool_bridge`; it does not exist. I wrote it against an earlier version that put the guard in `approval_manager`, deleted it when tach forced the guard into `tool_system.tool_hooks`, and never re-added it. The bridge half's only test change was a message string on `test_sync_tool_approval_execute_on_runtime_loop_fails_fast`, which already passed on base through the transport guard. So that half shipped with no new coverage and a PR body that claimed otherwise. Body corrected.

The other two blockers on that half are also right and both need real fixes, not touch-ups:

- **Policy evaluated twice.** `_approval_would_be_required` runs `evaluate_tool_approval`, then `request_tool_approval_for_call` runs it again. A stateful or scripted policy can answer False then True and walk straight into the path the guard exists to prevent. It also makes a script error return True and emit the sync-bridge message instead of the policy-failure one. Fix is to evaluate once and thread the result through.
- **`running_loop_is_sync_tool_bridge()` is the wrong predicate.** It answers "some other loop is blocked", not "I am that loop's bridge". Any unrelated loop — a worker, a test, another thread's `asyncio.run` — reports as a bridge and gets its approval declined. Should be a `ContextVar` set by `_run_deferred_result_from_sync`, which is exact and cheap.

Also fair: "Ask again so it runs asynchronously" is not actionable, since retrying does not make a `def` into an `async def`.

## Still open on this PR's own half

The recorded-decision-whose-edit-failed case, confirmed by two independent probes against this head:

```text
first sweep: complete, skipped_live_waiter=1
human decision: approved, edit failed
row remains=1
next sweep: failed=1, complete=false
```

`_resolved_card_event_ids` is added unconditionally, so redelivery is blocked forever. Pre-existing and listed as a follow-up, but worth being plain that merging this does not finish approval recovery.

One reviewer went further and argued the skip itself orphans the NULL-`card_event_id` row it was added for: pre-PR, `_identified_card` resent the frozen transaction, the homeserver deduplicated it, and `acknowledge_approval_card` repaired the row, so the waiter's later decision could land. Post-PR nothing repairs it, and the card stays visibly pending. The proposed remedy is narrow and does not resend: when `claimed.card_event_id is None` and a live waiter exists, re-attempt `acknowledge_approval_card(waiter.transaction_id, waiter.card_event_id, waiter.card_event)` from the guard, which already holds the waiter. I have not implemented or verified that yet; flagging it because it argues this PR trades one failure for another rather than removing one.

On the O(rows × waiters) scan: measured at 7 µs per lookup with 500 live waiters, so ~15 ms across a 10 000-row sweep. Not worth a second index and the consistency risk that comes with it.

PR #1822, opened 2026-08-12
fix: key pending-event lanes on the conversation, not the room

basnijholt:
## The problem

A pending tool approval in one thread made the agent stop answering **every** thread in that room. One production room spent 65 minutes in that state with five threads waiting; all five recorded a turn within two seconds of the approval finally being decided.

The pending-event worker keyed its lanes on `room_id`, so every conversation in a room ran behind whichever one was slowest. A turn parked on a human decision holds its lane for as long as the human takes, which is days by default.

## Why the obvious fix does not work

The first version of this PR keyed on `(room_id, event.thread_id)`. That changes nothing for the reported incident: `thread_id` is the `m.thread` relation and only that, so a root message records `None` and every root in a room lands back in one lane. The blocked events in production were all roots.

The conversation a root belongs to is the one it starts. That is the same thread root `MessageTarget.resolve` derives (`resolved_thread_id = thread_id or thread_start_root_event_id`), so the lane is `thread_id or event_id`.

## The invariant this turns on

The lane and the response lifecycle lock must name the same conversation:

```python
# response_lifecycle.py
def _thread_key(target: MessageTarget) -> tuple[str, str | None]:
    return (target.room_id, target.resolved_thread_id)
```

The lane decides what may run **concurrently**; the lock decides what must run **alone**. Keys that disagree either serialise a room that need not be, or start two turns that then contend for one lock and answer a thread out of receipt order. `test_a_lane_matches_the_thread_root_the_response_lock_uses` asserts the two agree rather than asserting the lane against itself.

That invariant is what produces the other two cases:

**Non-message events keep the room-wide lane.** A reaction, an edit, and a redaction also carry no `m.thread` relation, but they belong to whatever conversation their *target* is in, which the journal event does not record. Splitting them off would put a reaction in a lane beside the thread it annotates. `JournalEvent.names_its_own_conversation` draws that line.

**Room-scoped agents keep one lane per room.** An agent in room mode resolves every message to `resolved_thread_id = None`, so its lock already serialises the room by design. A lane per root would start those turns concurrently and let them take that single lock in arrival order. Only the bot knows its own thread mode, so it passes down the predicate it already had (`_room_scope_is_single_conversation`).

## What this does not fix

The thread holding the approval still parks until the decision arrives or the approval expires. This only stops it taking the rest of the room down with it. Releasing the parked thread itself is #1807 (durable continuations) or a shorter `timeout_days`.

## Tests

Written test-first; each new test was watched failing for the right reason before the change.

| Test | Guards |
|---|---|
| `test_a_parked_root_message_does_not_hold_another_root_message` | the incident: two roots, one parked |
| `test_a_lane_matches_the_thread_root_the_response_lock_uses` | lane key equals the lock's key |
| `test_a_room_scoped_agent_still_answers_its_room_in_one_order` | room mode stays serialised |
| `test_a_dispatchers_room_scope_reaches_the_lanes_it_owns` | the predicate is actually wired |

Four mutations run against the finished change; each was killed by exactly one distinct test:

| Mutation | Killed by |
|---|---|
| root falls back to the room lane | the two thread-scope tests |
| drop the non-message guard | `test_a_recovery_drain_runs_a_room_through_its_lane_not_beside_it` |
| ignore the room-scope predicate | `test_a_room_scoped_agent_still_answers_its_room_in_one_order` |
| drop the dispatcher wiring | `test_a_dispatchers_room_scope_reaches_the_lanes_it_owns` |

Existing tests that expressed "one room" with bare roots now express one thread, since the invariant they were written for is per-conversation.

`tests/test_journal_ingress.py`: 95 passed. Full suite diffed against merge base `15a2638ce`: **72 failures on the branch, 75 on baseline, zero new** (the rest are missing optional deps in this environment).

### DIFF
--- a/src/mindroom/bot.py
+++ b/src/mindroom/bot.py
@@ -673,6 +673,7 @@ class AgentBot:
             ),
             room_for_id=self._room_for_journal_event,
             on_persist_failure=self._record_dispatch_persist_failure,
+            room_is_one_conversation=self._room_scope_is_single_conversation,
             room_lifecycle_admission_enabled=lambda: (
                 self.agent_name == ROUTER_AGENT_NAME and self._first_sync_done and self._room_member_join_hooks_armed
             ),
--- a/src/mindroom/event_journal/models.py
+++ b/src/mindroom/event_journal/models.py
@@ -152,6 +152,19 @@ class JournalEvent:
     receipt_order: int
     semantic_consumer: SemanticConsumer | None = None
 
+    @property
+    def names_its_own_conversation(self) -> bool:
+        """Return whether this event says which conversation it belongs to.
+
+        A message does: ``thread_id`` is its thread, and a message without one
+        is the root of the thread it starts. Nothing else does. ``thread_id``
+        is the ``m.thread`` relation and only that, so a reaction, an edit, and
+        a redaction all record ``None`` while belonging to whatever
+        conversation the event they point at is in -- which is not recorded
+        here and cannot be inferred from here.
+        """
+        return self.kind is EventKind.MESSAGE
+
 
 # A non-empty ``__slots__`` is a TypeError on a subtype of a variable-length
 # built-in, so SLOT001 cannot be satisfied by a tuple that carries anything.
--- a/src/mindroom/journal_dispatch.py
+++ b/src/mindroom/journal_dispatch.py
@@ -41,7 +41,7 @@ from mindroom.matrix.journal_ingress import (
     projected_event,
 )
 from mindroom.matrix.media import MATRIX_MEDIA_EVENT_TYPES, MatrixMediaEvent
-from mindroom.pending_event_worker import PendingEventWorker
+from mindroom.pending_event_worker import PendingEventWorker, every_room_is_threaded
 
 if TYPE_CHECKING:
     from collections.abc import Awaitable, Callable
@@ -98,6 +98,11 @@ class JournalDispatcher:
     callbacks: JournalCallbacks
     room_for_id: Callable[[str], nio.MatrixRoom]
     on_persist_failure: Callable[[], None] | None = None
+    # Whether this agent reads a room as one conversation rather than a set of
+    # threads. Only the bot knows its own thread mode, and the worker this
+    # dispatcher builds needs it to key lanes the way the response lifecycle
+    # lock keys itself.
+    room_is_one_conversation: Callable[[str], bool] = every_room_is_threaded
     room_lifecycle_admission_enabled: Callable[[], bool] = lambda: False
     # Replaying a turn needs the agent fleet up, so the orchestrator releases
     # turn-backed replay separately from the rest of startup. Until it does,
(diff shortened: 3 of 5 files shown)


basnijholt:
Converting to draft. The mechanism and the regression attribution hold, but **this fix does not fix the incident**, and my test is why I did not notice.

`JournalEvent.thread_id` is raw Matrix relation data. A root message carries no thread relation, so it arrives as `''`/`None` and only becomes its own conversation later, when the canonical thread is resolved downstream — `conversation_resolver.py:359` (`fallback_root_event_id if event_info.can_be_thread_root`) feeding `turn_controller.py:859` (`resolved_thread_id=canonical_thread_id`). Reactions get their thread from interactive-question state, later still.

In the production incident every blocked event was a root:

```
blocking reaction $5807…   thread_id=''
blocked roots $AUE…, $29Kn…, $4qAZ…, $ll8…   thread_id='' (all four)
only genuinely threaded reply $Rsg…   thread_id='$29Kn…'
```

So `(room_id, JournalEvent.thread_id)` puts the blocker and all four roots in the same `(room, '')` lane. Sixty-five minutes of silence, unchanged.

My own log evidence said as much and I read past it: the `handled_turn_recorded thread_id=$AUE1wT0f…` values are the roots' *own event ids*, assigned downstream — not what the journal held.

**Why the test passed anyway.** `test_one_threads_parked_turn_does_not_hold_another_thread` builds both events with `text_event(..., thread_id="$thread-one" / "$thread-two")`, which sets an explicit `m.thread` relation. That is the one input shape where the raw journal thread id is already canonical. I picked the input that made the change look correct instead of the input from the incident I was fixing. Watching it fail proved only that *threaded replies* stop blocking each other — a real but much narrower case than the one reported.

What the change does still do: threaded replies in the same room no longer serialise behind each other, and `_failed_lanes` / `_lanes_with_more` follow the lane key. Neither helps the reported case.

Two ways forward, from the review:

1. **Hand interactive-response work to per-thread ownership durably, then release the room lane.** Targets the blocker directly rather than re-keying everything.
2. **Derive canonical lane keys asynchronously for every event kind**, roots and reactions included — the same resolution `MessageTarget` performs, done before lane assignment. Bigger, and it makes lane assignment async, which the collect path is not today.

Leaving this in draft rather than deleting it: the lane-key plumbing and the `_failed_lanes` / `_lanes_with_more` corrections are reusable by (2), and the test is worth keeping once it is rewritten around root messages so it actually reproduces the incident.

basnijholt:
Closing because the tool-approval failure this PR was created for has been superseded by #1807.

#1807 now persists a paused approval as a durable continuation, ends the live response lifecycle, releases typing and the conversation lock, and resumes later through managed response work. Human decision time therefore no longer keeps the `PendingEventWorker` callback or room lane occupied. The 65-minute room-wide outage described here cannot recur through a pending approval wait.

A broader question remains: `PendingEventWorker` still keys synchronous callback lanes by room, so any genuinely slow synchronous callback can cause cross-thread head-of-line blocking. That is now separate from tool approval waiting; normal responses and approval continuations are handed off instead of being awaited inside the journal callback.

This branch should not be revived as-is:

- It conflicts with current `main` after the continuation and interactive-dispatch changes.
- CI is red because `test_a_rooms_events_run_in_receipt_order` still asserts ordering between root messages that this change intentionally makes concurrent.
- `_WHOLE_ROOM_LANE` is another independent lane key, not a barrier across per-thread lanes, so it does not preserve reaction/edit/redaction ordering against message lanes.

If current-main evidence later shows slow synchronous callbacks causing material cross-thread latency, that should be handled in a fresh PR with an actual callback-level reproduction and explicit barrier semantics for room-wide events.

PR #1907, opened 2026-08-27
Settle ignored managed events before thread hydration

basnijholt:
## Summary

- Apply the existing managed-sender mention gate before conversation history hydration.
- Settle non-actionable managed events without scheduling a journal retry when related history is unavailable.
- Require every replayable source in a coalesced turn to belong to the classified requester before whole-turn settlement.
- Preserve dispatch for mentioned and explicitly exempt synthetic events, with the existing post-hydration guard retained as defense in depth.

## Regression coverage

One controller regression queues an unmentioned file event from a managed sender whose thread history returns M_NOT_FOUND. It verifies that the source is settled, never handed back for retry, and never reaches history reads, policy, or response execution.

A second regression builds an active-follow-up batch containing a human source followed by an unmentioned managed attachment. It verifies that unavailable history returns the entire mixed-requester batch for retry instead of settling the human source.

## Verification

- uv run pytest --no-cov (15,359 passed, 23 skipped)
- uv run tach check --dependencies --interfaces

## Summary by Sourcery

Settle non-actionable managed events before thread hydration while preserving valid dispatch and safe batch handling.

Bug Fixes:
- Settle unmentioned events from managed senders before conversation history hydration, preventing unavailable history from turning non-actionable events into retries.

Enhancements:
- Preserve dispatch for mentioned and explicitly exempt synthetic events while retaining the post-hydration policy check as a safeguard.
- Prevent suppression of mixed follow-up batches when replayable sources belong to different requesters.

Tests:
- Add regression coverage for managed attachments with unavailable thread history and mixed human/managed follow-up batches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of messages from managed senders when no agent is explicitly mentioned.
  * Ignored, unmentioned attachments are now settled before conversation history loads, preventing unnecessary processing and retries.
  * Improved mixed-source message handling so unrelated events are not incorrectly suppressed.
  * Preserved correct behavior for trusted relay and hook-originated messages.

* **Tests**
  * Added regression coverage for ignored attachments, mixed-source batches, and explicit agent mentions in hook-generated messages.

### DIFF
--- a/src/mindroom/conversation_resolver.py
+++ b/src/mindroom/conversation_resolver.py
@@ -228,6 +228,14 @@ class DispatchContextResult:
 
 
 @dataclass(frozen=True)
+class _PreHydrationPolicyFacts:
+    """Policy facts available without resolving thread history."""
+
+    origin: TurnOrigin
+    am_i_mentioned: bool
+
+
+@dataclass(frozen=True)
 class ConversationResolverDeps:
     """Explicit collaborators for conversation resolution."""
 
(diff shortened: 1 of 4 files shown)


basnijholt:
Fixed in 15e289025. Whole-turn managed-sender suppression now requires every replayable source to be attributable to the classified requester. The new active-follow-up regression combines a human source with a managed primary attachment and verifies that unavailable history returns the mixed batch for retry instead of settling it.

Issue #1924, opened 2026-08-31
[feature][matrix] Deliver oversized responses as lossless Markdown segments


lizujie:
## Goal

Deliver oversized Matrix responses as complete, ordered, rich-text messages so the full Markdown response remains visible and rendered in Element.

This supersedes the earlier proposal that only kept a truncated rich-text preview next to a `message-content.json` sidecar. A sidecar may remain as a recovery/history mechanism, but it must not be the only user-visible representation of the answer.

## Background

With `defaults.enable_streaming: true`, a long Markdown response can render correctly during streaming, but the terminal edit may exceed the Matrix event limit. The current oversized path then falls back to a JSON sidecar. Element does not automatically hydrate this custom sidecar into the original rich-text message, so the final visible answer is incomplete or loses Markdown rendering.

## Steps to reproduce

1. Enable streaming for Matrix responses.
2. In a MindRoom Matrix room, ask an agent for a response long enough to exceed one Matrix event.
3. Observe the in-progress streaming update.
4. Wait for the response to complete and inspect the final events in Element.

## Expected behavior

When a response is too large for one Matrix event:

- The complete response is split into multiple ordered `m.room.message` events.
- Every segment keeps `format: org.matrix.custom.html` and `formatted_body` when the source is Markdown.
- If the first event is a final `m.replace` of a streaming placeholder, the remaining segments are sent as normal rich-text continuation messages in the same room; when a thread exists, they remain in that thread.
- Concatenating the segment bodies reproduces the original response exactly, without a visible truncation marker such as `[Message continues in attached file]`.
- Each event remains below the Matrix homeserver size limit, including the event metadata and (where applicable) encryption overhead.
- A sidecar can still be retained for local recovery, but it is not the primary visible content.

## Actual behavior

The terminal oversized edit is converted to a JSON sidecar event:

- `m.new_content.msgtype` becomes `m.file`
- The attachment is named `message-content.json`
- The visible outer content and `m.new_content` no longer contain the rich-text fields
- Element does not automatically fetch and render the full custom sidecar

The result is a truncated or plain-text-looking final response even when the streaming preview rendered correctly.

## Proposed implementation

Prepare the final Matrix payload before the durable delivery row is frozen:

1. Detect an oversized text/edit payload before the generic large-message sidecar fallback.
2. Split the body at paragraph/line boundaries where possible, otherwise at Unicode character boundaries, preserving every character.
[...]

PR #1927, opened 2026-08-31
Deliver oversized Matrix responses as lossless Markdown segments

lizujie:
## Summary

Addresses #1924.

Oversized Matrix text responses are no longer forced onto the truncated-preview + JSON sidecar path. When enabled, the delivery gateway splits an oversized response into multiple complete, ordered, rich-text `m.room.message` events so the full Markdown answer stays visible and rendered in any Matrix client — no sidecar hydration required.

The behavior is gated behind a new opt-in config switch and defaults to the existing sidecar behavior:

```yaml
defaults:
  large_message_strategy: split   # default: sidecar
```

## How it works

- **`matrix/segmented_messages.py`** — splits the response body at paragraph/line boundaries (falling back to Unicode boundaries), preserving every character; concatenating segment bodies reproduces the original response exactly. Every segment keeps `format: org.matrix.custom.html` with a freshly rendered `formatted_body`. Internal streaming baggage (`io.mindroom.tool_trace`, `io.mindroom.visible_body`) is stripped from the visible first event so a long tool trace alone cannot force the answer into a sidecar.
- **First segment vs continuations** — the first segment stays a final `m.replace` of the streaming placeholder; continuations are normal rich-text messages that remain in the same thread when one exists.
- **Conservative per-event budgets** — 46 KB plaintext / 22 KB edit / 30 KB encrypted / 16 KB encrypted edit (estimated event size including a 2 KB margin), deliberately below the sidecar thresholds so every segment passes through `large_messages` unchanged.
- **Durable recovery** — continuation payloads are frozen into the local outbox record (`io.mindroom.matrix_segment_payloads`, never sent to Matrix), and each continuation gets a deterministic `uuid5` transaction ID derived from the turn, so retries after a crash/restart are idempotent and cannot post duplicates.
- **Sidecar remains the fallback** — non-text payloads, or metadata that alone exceeds the budget, still use the existing `large_messages` sidecar path; streaming in-progress edits are untouched.

## Commits

1. `Split oversized Matrix text responses into lossless segmented messages` — segmentation module, delivery-gateway integration, regression tests.
2. `Add opt-in large_message_strategy config for segmented Matrix responses` — the `defaults.large_message_strategy` switch (`sidecar` default, hot-reload aware), frontend type preservation, sidecar-mode regression test.

## Testing

- `tests/test_segmented_messages.py` — exact body preservation, rich-text fields, per-segment size budgets, thread relations, tool-trace stripping.
- `tests/test_response_delivery_gateway.py` — end-to-end outbox flow: frozen primary + continuations fit the hard limit, `room_send` receives `[primary, *continuations]`, no `m.file`/upload in split mode; sidecar mode asserts the legacy `m.file` + single upload behavior.
- `tests/test_config_manager_consolidated.py` — config parsing: `sidecar` default, `split` accepted, invalid values rejected.
- Full `pre-commit run --all-files` passes (ruff, ty, tach, privata, TypeScript check).

## Notes for reviewers

- With the default `sidecar`, observable behavior is unchanged from `main`; the issue's acceptance criteria are met with `large_message_strategy: split`. Happy to flip the default if preferred.
- `frontend/src/types/config.ts` includes a prettier reformat (double→single quotes) enforced by the repo's own hook on any touch of that file; the only semantic change is the added optional `large_message_strategy` field.

## Summary by Sourcery

Enable opt-in lossless segmentation of oversized Matrix Markdown responses while preserving the existing sidecar behavior by default.

New Features:
- Add an opt-in Matrix delivery strategy that splits oversized Markdown responses into ordered, lossless rich-text message segments.
- Preserve threading, mentions, formatting, and complete fenced code blocks across segmented Matrix responses.

Bug Fixes:
- Make segmented response delivery durable and crash-safe by persisting continuation payloads locally and recovering only missing events without duplicating identical segments.

Enhancements:
- Retain the existing sidecar upload behavior as the default fallback for unsupported or unsplittable oversized payloads.

Documentation:
- Document the new large_message_strategy configuration option and its sidecar default.

Tests:
- Add coverage for lossless segmentation, formatting, threading, mentions, code fences, size budgets, configuration validation, sidecar fallback, and recovery behavior.

## Summary by CodeRabbit

* **New Features**
  * Added configurable handling for oversized responses.
  * Responses can be delivered as an attachment preview or split into lossless continuation messages.
  * Split messages preserve Markdown formatting, edits, threading, mentions, and safe Unicode boundaries.
  * Added configuration support for selecting the delivery strategy.

* **Bug Fixes**
  * Improved recovery to resend only missing continuation messages.
  * Prevented unsafe splitting of fenced code blocks and misidentified mentions.

* **Documentation**
  * Updated the example configuration with the new oversized-message setting.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -287,6 +287,7 @@ defaults:
   tools: [scheduler]
   markdown: true
   enable_streaming: true
+  large_message_strategy: sidecar
 
 memory:
   backend: mem0
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -157,6 +157,7 @@ Matrix sync callback
 | `credentials.py` | Unified credential management (CredentialsManager) |
 | `matrix/` | Matrix protocol integration (client, users, rooms, presence, provisioning, message formatting) |
 | `matrix/large_messages.py` | Large-message sidecar storage and retrieval for oversized Matrix payloads |
+| `matrix/segmented_messages.py` | Lossless splitting of oversized text responses into ordered rich-text events (`defaults.large_message_strategy: split`) |
 | `matrix/sync_checkpoint_trust.py` | Sync-checkpoint persistence and the journal generation a checkpoint is certified against |
 | `matrix/sync_continuity.py` | Crash-atomic checkpoint and pending join-fence persistence |
 | `matrix/journal_ingress.py` | The boundary where Matrix events become durable facts; nio provenance decides actionable vs context-only |
--- a/docs/architecture/matrix.md
+++ b/docs/architecture/matrix.md
@@ -169,6 +169,13 @@ Messages exceeding the 64KB Matrix event limit are automatically handled by `pre
 - Preview event is compact (for example no inline `io.mindroom.tool_trace`), while the sidecar preserves full content fidelity
 - Encrypted rooms: sidecar JSON is encrypted before upload (`message-content.json.enc`)
 
+With `defaults.large_message_strategy: split`, an oversized final text response is instead delivered as several complete rich-text events by `segment_matrix_content()` in `matrix/segmented_messages.py`.
+The body is cut at paragraph or line boundaries, never inside a fenced code block, and concatenating the segment bodies reproduces the original exactly.
+The first segment stays a final `m.replace` of the streaming placeholder when there is one; continuations are plain messages that stay in the thread when there is one.
+Every segment is rendered as standalone Markdown with `m.mentions` attached to the segment whose body carries the mention.
+Continuation payloads are frozen in the local outbox row and sent under deterministic transaction IDs, so a retry or restart resends only the segments the room does not already hold.
+Non-text payloads, metadata that alone exceeds the budget, and a single code fence larger than one event still use the sidecar path.
+
 ## Response Tracking
 
 Duplicate responses are prevented at two durable layers, both in `tracking/event_journal.db` under `mindroom_data/`.
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -390,6 +390,7 @@ defaults:
     - scheduler                    # Plain string or single-key dict with inline config overrides
   markdown: true                   # Default: true
   enable_streaming: true           # Default: true (stream responses via message edits)
+  large_message_strategy: sidecar  # Default: sidecar (or split: deliver oversized responses as lossless segmented messages)
   streaming:
     update_interval: 5.0           # Default: 5.0 (steady-state seconds between streamed edits)
     min_update_interval: 0.5       # Default: 0.5 (fast-start seconds between early edits)
--- a/docs/streaming.md
+++ b/docs/streaming.md
@@ -156,6 +156,7 @@ If an error occurs during streaming, the message is finalized with:
 ## Large Streamed Messages
 
 If a streamed response exceeds the Matrix event size limit (55KB for new messages, 27KB for edits), the large message system automatically uploads a JSON sidecar and includes a preview in the event body.
+With `defaults.large_message_strategy: split`, the final edit is instead delivered as several complete rich-text messages so the whole Markdown answer stays visible without a sidecar.
 See [Matrix Integration — Large Messages](architecture/matrix.md#large-messages) for details.
 
 ## Visibility Toggles
--- a/frontend/src/types/config.ts
+++ b/frontend/src/types/config.ts
@@ -1,18 +1,15 @@
-import type { PROVIDERS } from '@/lib/providers';
+import type { PROVIDERS } from "@/lib/providers";
 
 export type ProviderType = keyof typeof PROVIDERS;
-export type MemoryBackend = 'mem0' | 'file' | 'none';
-export type MemorySearchMode = 'keyword' | 'semantic';
-export type WorkerScope = 'shared' | 'user' | 'user_agent';
-export type PrivateWorkerScope = Exclude<WorkerScope, 'shared'>;
+export type MemoryBackend = "mem0" | "file" | "none";
+export type MemorySearchMode = "keyword" | "semantic";
+export type WorkerScope = "shared" | "user" | "user_agent";
+export type PrivateWorkerScope = Exclude<WorkerScope, "shared">;
 export type AgentPolicySource =
-  | 'private.per'
-  | 'agent.worker_scope'
-  | 'defaults.worker_scope'
-  | 'unscoped';
+  "private.per" | "agent.worker_scope" | "defaults.worker_scope" | "unscoped";
 export type AgentPoliciesByAgent = Record<string, AgentPolicy>;
-export const DEFAULT_PRIVATE_KNOWLEDGE_PATH = 'memory';
-export const SHARED_CONTEXT_FILE_PLACEHOLDER = 'SOUL.md';
+export const DEFAULT_PRIVATE_KNOWLEDGE_PATH = "memory";
+export const SHARED_CONTEXT_FILE_PLACEHOLDER = "SOUL.md";
 
 export interface ModelConfig {
   provider: ProviderType;
@@ -82,7 +79,7 @@ export interface KnowledgeGitConfig {
 }
 
 export interface KnowledgeBaseConfig {
-  mode?: 'semantic' | 'files';
+  mode?: "semantic" | "files";
   description?: string;
   path: string;
   watch: boolean;
@@ -109,10 +106,10 @@ export interface AgentPrivateConfig {
   knowledge?: AgentPrivateKnowledgeConfig | null;
 }
 
-export type LearningMode = 'always' | 'agentic';
-export type CultureMode = 'automatic' | 'agentic' | 'manual';
+export type LearningMode = "always" | "agentic";
+export type CultureMode = "automatic" | "agentic" | "manual";
 
-export type ThreadMode = 'thread' | 'room';
+export type ThreadMode = "thread" | "room";
 
 export interface ResponderAccessConfig {
   current_room_members?: boolean;
@@ -131,30 +128,34 @@ export interface CompactionConfig {
   fallback_model?: string | null;
 }
 
-const DEFAULT_INHERITED_TOOLS = ['scheduler'] as const;
+const DEFAULT_INHERITED_TOOLS = ["scheduler"] as const;
 
 // Every authored compaction field except the model names, which carry their own
 // clear semantics. Keep in sync with CompactionOverrideConfig in the backend:
 // authoring any of these marks the override as authored, which the backend
 // resolves to enabled compaction.
 const COMPACTION_NON_MODEL_FIELDS: readonly (keyof CompactionConfig)[] = [
-  'enabled',
-  'threshold_tokens',
-  'threshold_percent',
-  'replay_window_tokens',
-  'reserve_tokens',
-  'timeout_seconds',
+  "enabled",
+  "threshold_tokens",
+  "threshold_percent",
+  "replay_window_tokens",
+  "reserve_tokens",
+  "timeout_seconds",
 ];
 
-function hasAuthoredNonModelCompactionField(compaction: CompactionConfig): boolean {
-  return COMPACTION_NON_MODEL_FIELDS.some(field => compaction[field] !== undefined);
+function hasAuthoredNonModelCompactionField(
+  compaction: CompactionConfig,
+): boolean {
+  return COMPACTION_NON_MODEL_FIELDS.some(
+    (field) => compaction[field] !== undefined,
+  );
 }
 
 function isPureCompactionModelClear(compaction: CompactionConfig): boolean {
   const modelFields = [compaction.model, compaction.fallback_model];
   return (
-    modelFields.some(value => value === null) &&
-    modelFields.every(value => value === null || value === undefined) &&
+    modelFields.some((value) => value === null) &&
+    modelFields.every((value) => value === null || value === undefined) &&
     !hasAuthoredNonModelCompactionField(compaction)
   );
 }
@@ -168,7 +169,7 @@ function isEmptyCompactionOverride(compaction: CompactionConfig): boolean {
 }
 
 function resolveAuthoredCompactionEnabled(
-  compaction: CompactionConfig | null | undefined
+  compaction: CompactionConfig | null | undefined,
 ): boolean {
   if (compaction === null) {
     return false;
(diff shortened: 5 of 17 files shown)


basnijholt:
Very cool! First real contribution! I ran it through AI review which gave me:

---

❌ **CHANGES REQUIRED**

1. Cross-device recovery can permanently lose continuations.  
[delivery_gateway.py:958](/work/worktrees/mindroom-pr-1927-review/src/mindroom/delivery_gateway.py:958) searches only for the primary payload, and [matrix_delivery.py:346](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix_delivery.py:346) acknowledges the entire row immediately when that event is found.  
After a crash between segment sends, a new device therefore settles the delivery without calling the continuation loop at [delivery_gateway.py:817](/work/worktrees/mindroom-pr-1927-review/src/mindroom/delivery_gateway.py:817).  
A direct probe confirmed one primary reconciliation, zero continuation sends, and a successful acknowledgement.

2. Long fenced code blocks lose Markdown formatting.  
[segmented_messages.py:221](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix/segmented_messages.py:221) splits without tracking Markdown state, while [segmented_messages.py:104](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix/segmented_messages.py:104) renders each chunk independently.  
A five-segment fenced-code probe rendered the first segment as code but not the second, violating the acceptance criterion that Markdown remain rendered in every segment.

3. Matrix mention links are discarded.  
Mention formatting intentionally creates link-bearing HTML separately from plaintext at [mentions.py:462](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix/mentions.py:462).  
The segmenter discards that HTML and regenerates it from plaintext at [segmented_messages.py:102](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix/segmented_messages.py:102).  
The probe retained `m.mentions` but lost the `matrix.to` hyperlink; mentions occurring in later chunks also leave their metadata on the wrong event.

4. Room-mode continuations break visible-response recovery.  
[segmented_messages.py:153](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix/segmented_messages.py:153) copies the primary standalone reply relation onto every continuation.  
The history predicate at [room_history_reads.py:269](/work/worktrees/mindroom-pr-1927-review/src/mindroom/matrix/room_history_reads.py:269) consequently treats every segment as an independent response, after which [visible_response_reconciliation.py:120](/work/worktrees/mindroom-pr-1927-review/src/mindroom/visible_response_reconciliation.py:120) raises for multiple responses.  
A three-segment probe produced three matching response events.

5. The checked-in frontend file fails the repository formatter hook.  
On the exact PR commit, `uv run pre-commit run prettier --files frontend/src/types/config.ts --verbose` rewrites [config.ts:1](/work/worktrees/mindroom-pr-1927-review/frontend/src/types/config.ts:1) and exits unsuccessfully.  
The unrelated whole-file reformat should be removed or made consistent with the actual hook.

6. The public configuration setting is undocumented.  
The field is introduced at [models.py:364](/work/worktrees/mindroom-pr-1927-review/src/mindroom/config/models.py:364), but the user-facing defaults documentation at [docs/configuration/index.md:387](/work/worktrees/mindroom-pr-1927-review/docs/configuration/index.md:387) does not mention it.  
Adding it only to `CLAUDE.md` does not document how operators enable the feature.

Verification: 383 tests passed and Tach passed.  
The full pre-commit run failed only because Prettier modified `frontend/src/types/config.ts`.  
No live Matrix test was run because deterministic probes already expose merge-blocking failures.

Automated-review feedback was treated as untrusted and independently verified.  
The separate partial-restart testing comment is covered by finding 1; the JSDoc request was rejected because it targets pre-existing exports affected only by formatter churn and is not a repository rule.

The requested `caveman` skill was unavailable in this session.

---


Haven't had the chance to look at it in detail myself.

lizujie:
Thanks for the quick review! One note on timing: findings 1-4 match the pre-fix revision — the bot reviewers (Qodo, CodeRabbit) flagged the same issues earlier, and fixes were pushed before your run. Current HEAD is 05740d3. Point by point:

1. **Cross-device recovery losing continuations** — fixed in 4e3192c. Adoption now matches every continuation by its exact frozen content and sends only the missing ones (`_deliver_missing_continuations`), each under its stable derived transaction ID; any failure leaves the row unacknowledged so the next recovery pass resumes there. A regression test covers exactly your probe: crash after the primary + one continuation, new device sends only the missing segments, then acknowledges.

2. **Fenced code blocks** — fixed in 4e3192c and hardened in 185ac55 / ca36cca. Cuts move ahead of an unclosed fence, closers follow CommonMark (same marker character, at least the opening length, trailing whitespace only, same block-quote depth), and a fence spanning the whole budget falls back to the sidecar.

3. **Mention links** — fixed in 4e3192c: the pill anchors from the source `formatted_body` are carried into the rendered HTML of whichever segment contains the mention (single longest-first regex pass since 185ac55, so prefix-overlapping user IDs can't nest). Your metadata sub-point is addressed in 05740d3: each segment now carries `m.mentions` for just the user IDs present in its own body.

4. **Room-mode continuation relations** — fixed in 4e3192c. Continuations always take the thread-fallback form (or no relation outside threads), so the history predicate counts exactly one visible response per segmented turn; regression tests pin both the room-reply and genuine-thread-reply cases.

5. **Prettier churn** — confirmed real, and embarrassing: that whole-file reformat came from a stale hook environment on my side. Fixed in 56fc00e — the file is back to main's formatting with only the new field added, so the diff against main is a single line and `uv run pre-commit run prettier --files frontend/src/types/config.ts` passes on a clean install.

6. **Missing docs** — `defaults.large_message_strategy` is now documented in `docs/configuration/index.md` (b366677).

Verification on HEAD: the touched test files pass (236 tests, including the new recovery/segmentation regressions), and the full pre-commit suite passes. Could you re-run your review against 05740d3? The CI workflows are also still waiting on an approval run whenever you get a chance.

basnijholt (dismissed):
Disclaimer: I personally still haven’t had time to properly review this PR. The findings below come from an assisted rereview and should be treated as preliminary until I can review it myself.

I rechecked the latest head (`05740d359`) and the current review threads. The earlier room-relation, ordinary fence, basic mention-pill, formatting, and documentation issues appear addressed, but these blockers remain:

1. **Cross-device recovery cannot reconcile repeated identical continuations.** `_deliver_missing_continuations()` checks each expected continuation by exact-content existence independently. Homogeneous long responses can produce many byte-identical continuation payloads, so one observed event can satisfy several positions, while multiple copies can make the exact-match helper reject recovery. This can acknowledge an incomplete response or leave it unrecoverable. Please make reconciliation position/index- or multiplicity-aware and add a repeated-payload changed-device regression.

2. **Segmented mention metadata is not Matrix-correct.** `_chunk_mentions()` drops `m.mentions.room`. For edits, `_edit_content_chunk()` also replaces the top-level revision delta with mentions derived from `m.new_content`, potentially re-notifying existing recipients. Preserve room mentions, keep the edit delta at the outer level, and keep the resolved final set under `m.new_content`.

3. **Mention-pill restoration rewrites literal content.** `_render_segment_html()` substitutes user IDs across the rendered HTML, which can insert a pill inside `<code>` or other non-mention text. Its substring test can also attribute a shorter prefix-overlapping user ID. Mention preservation needs to be source-position-aware.

4. **Valid indented quoted fences can still be split.** The fence matcher handles only one optional whitespace after `>`, but CommonMark accepts a fence indented up to three spaces within the quote, such as `>   ```python`. An oversized block using that form can be split between its opener and closer and render as prose.

Verification on the current head: 294 focused/adjacent tests pass, as do the full pre-commit suite and Tach. The failures above were reproduced with direct payload probes, and the current tests do not cover them.

lizujie:
Thanks for the recheck — all four blockers were valid and are now fixed on HEAD (`0277cd2`). Point by point:

1. **Repeated identical continuations** — confirmed; my own first recovery test actually produced them (a homogeneous 100 KB body splits into byte-identical 21 KB chunks), so this one stings. Reconciliation is now multiplicity-aware: `missing_outbox_delivery_copy_indices_via_room_messages` counts exact copies per payload and consumes them as credits across the expected positions, so one observed event satisfies exactly one position and the multi-copy case no longer trips the exact-match guard. (`f403825`) The repeated-payload changed-device regression you asked for runs against the real store and the real scan: old device delivers the primary + one of two identical continuations, new device resends exactly the second duplicate and the tail, each under its stable derived transaction ID.

2. **Mention metadata** — `_chunk_mentions` now preserves `room: true`, attributed to the segment whose body carries `@room` (room-only and combined cases covered). For edits, the outer envelope keeps the source edit's own `m.mentions` — the revision delta is no longer re-derived, so no re-notification risk — while the resolved per-chunk set lives under `m.new_content`. (`0277cd2`)

3. **Literal-content rewriting** — pill restoration is now boundary-aware and code-excluding: matches require ID-boundary lookarounds (so `@alice:localhost` no longer attributes inside `@alice:localhost2` or inside URLs), and `<code>` regions are skipped entirely. A user ID appearing as literal text stays literal. (`0277cd2`)

4. **Indented quoted fences** — the quote prefix now allows up to three spaces after each `>` marker per CommonMark, so `>   ```python` fences are recognized, with regressions at both the helper level and across a real split boundary. (`0277cd2`)

Verification on `0277cd2`: 277 tests across the touched delivery/segmentation/reconciliation files pass (including the new regressions), the full pre-commit suite passes, and `tach.toml` was updated in the same commits since the scan function crosses a module boundary.

basnijholt:
Pushed `6183d5cea` addressing the review findings directly:

- **Event-loop stall / over-fragmentation**: splitter now shrinks proportionally to the measured overshoot (1–3 renders per chunk instead of bisecting with 3 renders per probe) and sizes continuations against the plain budget instead of the edit budget. Measured on an edit in a plaintext room: 51 KB body went from 14 messages / 2.4 s to 4 messages / ~170 ms; 115 KB from 41 / 8.1 s to 8 / ~390 ms. The tail is no longer cut at a paragraph boundary when it already fits.
- **Edit fallback `formatted_body`** was `markdown_to_html("* " + body)`, which rendered the first heading inside `<ul><li>`; it now reuses the replacement HTML like `_wrap_large_edit`.
- **DRY**: `large_messages.calculate_event_size` / `is_edit_message` are public and reused; the two metadata copiers are one; dead `_split_body` builder parameter and per-line regex compile removed; segments no longer get `matrix_delivery_payload` re-applied (they already copy the identity from `wire_content`); dropped the `LargeMessageStrategy` annotation import.
- **Corrupt outbox result** (`io.mindroom.matrix_segment_payloads` not a list) now raises instead of silently acknowledging a row with continuations never sent.
- **Tests**: `_gateway()` defaults to the production `sidecar` strategy again and split tests opt in explicitly; restored the double-prepare and definitive-refusal sidecar regressions from `main`. Edit-segment test now asserts first ≤ edit budget, continuations ≤ plain budget.
- **Docs**: module table row in `CLAUDE.md`, split-strategy paragraph in `docs/architecture/matrix.md`, note in `docs/streaming.md`.

Verified: ruff, ty, tach clean; `tests/test_segmented_messages.py`, `tests/test_response_delivery_gateway.py`, `tests/test_large_messages.py`, `tests/test_config_manager_consolidated.py` pass (postgres-backed params skipped locally, no `psycopg`).

basnijholt:
Second pass at `6183d5cea`, pushed `697d88727`:

- `_send_claimed` / `_send_continuation` duplicated the refusal-to-exception mapping; both now go through one `_send_frozen`, the primary being the first frozen event.
- `_result_with_segment_payloads` takes the prepared union, so the two enqueue sites no longer repeat the `() if isinstance(prepared, MatrixDeliveryFailure) else ...` ternary.
- In `_count_delivery_copies_via_room_messages` the page-bound check could only ever raise (the loop already exits when everything is found), so the `if ...: break` around it was unreachable; `Counter` replaces the hand-rolled multiplicity dicts.
- Dropped the `split` flag from `test_delivery_identity_is_included_in_the_validated_event_size`, whose payload never segments.

ruff, ty, tach clean; segmented/gateway/large-message/room-history tests pass. Remaining PR reads clean to me: ✅ approve once CI is green.

basnijholt:
Live test at `697d88727` against local Synapse (`just local-matrix-up`), isolated backend (`MINDROOM_NAMESPACE=livepr1927`, `--api-port 9876`), deterministic OpenAI-compatible stub returning a 58,644-byte Markdown body (120 sections, lists, links, a 200-line fenced `python` block, `END-OF-LONGTEST` marker), `enable_streaming: true`.

**`large_message_strategy: split`** — room `!gaponUqjhhKRvutfzU:localhost`, thread root `$qDjUjcFG9r8FIRAYo2-8AXjxdTZGxeY8aaaB-kur1dk`:

- 1 placeholder + final `m.replace` edit `$jDgESMlHYlHF7wt4qD-l3gvLcWlfo33-5dCmrjRudwA`, 5 continuations, all `m.text` with `format: org.matrix.custom.html` + `formatted_body`.
- Segment body bytes `[3648, 17474, 13547, 11833, 6104, 6038]`; canonical event JSON bytes `[18053, 39294, 30711, 26959, 14376, 39239]` (all under the 46 KB plain / 22 KB edit budgets).
- Concatenated bodies `==` stub body exactly (58,644 chars). No `io.mindroom.long_text`, no `url`/`file`, no `[Message continues …]` marker.
- Fenced block lands whole in one segment; every segment fence-balanced.
- Continuation relation: `{"rel_type": "m.thread", "event_id": <root>, "is_falling_back": true, "m.in_reply_to": {"event_id": <placeholder>}}`.
- Outer edit fallback: body `* # Long answer…`, `formatted_body` starts `<h1>Long answer</h1>` (no bullet-list wrapping).
- Backend log: `Prepared lossless Matrix response segmentation … continuation_count=5`; no errors.

**Hot reload to `sidecar`** (edited `config.yaml`, `Configuration file changed; queueing hot reload`), same prompt: final edit `m.new_content.msgtype == "m.file"`, `io.mindroom.long_text = {version: 2, encoding: matrix_event_content_json, original_event_size: 310153, preview_size: 22000, is_complete_content: true}`, body ends with `[Message continues in attached file]`, 0 continuations. Default path unchanged, strategy switch takes effect without restart.

Not exercised live: crash-between-segments recovery (covered by `test_recovery_from_a_new_device_sends_only_the_missing_continuations` and `test_identical_continuations_reconcile_by_copy_count`), encrypted rooms.

basnijholt:
Thanks a lot for implementing this, @lizujie! Really nice work, especially the durable continuation handling and the fence-aware splitting.

I pushed a couple more fixes directly to the branch before merging (segment sizing per role, single render per chunk, the edit fallback HTML, and some deduplication in the gateway), then squash-merged. Live-tested both strategies against a local Synapse and everything behaved as described in the issue.

I'm keeping `sidecar` as the default for now while we dogfood `split`. One thing I noticed for a possible follow-up: when a user edits their message and the answer is regenerated, only the primary event gets rewritten, so continuations from the previous answer stay in the thread. Happy to discuss if you feel like picking that up.

PR #1954, opened 2026-09-04
Upstream thread-export plugin as per-agent thread_exports

basnijholt:
## Summary

Upstreams [`thread-export-plugin`](https://github.com/mindroom-ai/thread-export-plugin) into MindRoom as a per-agent `thread_exports` setting.
The plugin already delegated every export to `mindroom.thread_export`; what it added was trigger plumbing, target resolution, and workarounds for the plugin boundary (hot-reload guards, a private-root cache with revision counters, fresh Matrix logins and journal opens per pass, a private event loop in a worker thread). All of that goes away.

```yaml
agents:
  code:
    thread_exports: true            # defaults
  research:
    thread_exports:
      invited_rooms: false          # config rooms only
      private_room_scope: owner     # private agents only
```

## What changed

- **Config**: `AgentConfig.thread_exports: AgentThreadExportConfig | None`, with `true`/`false` shorthand.
- **Trigger**: `JournalIngress` gains `on_room_activity`, fired once per newly admitted `MESSAGE`/`MEDIA`/`REDACTION`/`ROOM_LIFECYCLE` event, so a member leaving re-evaluates the room's exports (actionable or context-only). This replaces the plugin's `message:received` + `message:after_response` hooks and also covers edits, redactions, and messages a bot ignores. `JournalDispatcher` passes it through; `AgentBot` takes a `room_activity_observer`.
- **Runner**: `thread_export/workspace_sync.py` debounces activity (2 s) into single-flight passes (one pass over the exact set of dirty room IDs), resolves shared and private targets, clears exports of agents that no longer enable them, and reads rooms through the live bots' clients and journal principals. Private instances are validated against their core identity record every pass; no cache.
- **Lifecycle**: the orchestrator owns one always-running `WorkspaceThreadExportRunner`: `start()` on the first support-service sync, `queue_full_pass()` after bots start and after every config reload, `stop()` at shutdown. A full pass with no enabled agent is the cleanup sweep, so there is no on/off state machine. A bot that is not running (router included) makes its rooms unreadable for that pass; the post-start or post-reload full pass catches up.
- **Private instances**: `private_instances_for_agent` in `private_instance_identity_store.py` enumerates instance roots and validates ownership (record ↔ requester ↔ directory ↔ current `private.per`). The runner only maps instances to workspace targets. `private_instance_state_root_for_requester` had no remaining callers and is removed.
- **Service split**: `export_threads_to_sources` is the in-process path (already-authenticated `ThreadExportSource`s). `export_threads_to_targets_once` keeps the CLI behaviour (login per account group, open/bind journal) and delegates to the same body.
- **Storage I/O off the loop**: thread writes, room reconciliation, index writes, retractions, target validation, and full-pass reconciliation now run via `asyncio.to_thread`. The plugin worked around this with a private loop in a thread; a live client cannot leave its loop, so the blocking layer moves instead. CLI gets the same path.
- **Removed**: `clear_thread_export_root(should_clear=...)` (plugin-only hook-liveness predicate) and its race test.
- **Docs**: `docs/configuration/agents.md` gets a Thread Exports section (layout, private instances, semantic-search recipe); `docs/cli.md` cross-references it.

## Not carried over

- `debounce_seconds` setting (constant 2 s).
- Hot-reload bookkeeping (`_runner_tasks`, shared pass lock, `is_active` threading).
- Deleting exports for an agent whose Matrix account is missing; a bot that has not started simply contributes nothing to the pass and the work stays queued.

## Migration

Remove the plugin from `plugins:` and set `thread_exports` on the agents that were in the plugin's `agents` list, in the same config edit. `invited_rooms` and `private_room_scope` keep their names. Existing `<workspace>/thread_exports/` trees are adopted as-is (same writer, same paths, same ownership marker); unchanged files are not rewritten.

The plugin is incompatible with this change, not merely redundant: it imports `private_instance_state_root_for_requester`, removed here with its last caller, so a config that still lists the plugin fails validation and MindRoom refuses to start or reload with it.

Archive PR for the plugin: mindroom-ai/thread-export-plugin#12

## CI ordering

`.github/plugin-fleet.json` no longer lists `thread-export-plugin`, so the `Build plugin matrix` drift check stays red until the plugin repository is archived on GitHub (the drift check compares the registry against non-archived `*-plugin` org repos). Archive the plugin repo right before or after merging this PR; the plugin's own compatibility check would fail against this branch anyway, because it drops the plugin-only `should_clear` hook.

## Testing

- `uv run pytest`: 15632 passed, 22 skipped.
- New: `tests/test_thread_export_workspace_sync.py`, `tests/test_agent_thread_export_config.py`, `private_instances_for_agent` cases in `tests/test_private_instance_identity.py`, `tests/test_bot_room_activity.py`, room-activity cases in `tests/test_journal_ingress.py`, source-path cases in `tests/test_thread_export_service.py`.
- `uv run tach check --dependencies --interfaces` clean. `pre-commit` clean except the pre-existing `privata` finding on `LargeMessageStrategy` in `config/models.py` (present on `main`).


basnijholt:
Review triage (pushed in the two latest commits):

- **Final disable leaves exports** (qodo) — fixed. `ThreadExportRuntimeCoordinator.sync()` now clears every agent's export tree when no agent enables `thread_exports` any more, since no runner is left to do it.
- **Crashed passes lose work** (qodo) — fixed. Pending rooms and the full-pass flag are restored after an exception; the next trigger retries.
- **Deferred requeue never wakes the loop** (CodeRabbit) — fixed. A pass that finds the router not running re-arms the loop, so the debounce becomes the retry interval. Router readiness is now checked before any disk work, so a deferred pass is free.
- **Room selection on the event loop** (CodeRabbit nit) — fixed, `_select_rooms` runs via `asyncio.to_thread`; it is also one call per pass now, not one per dirty room.
- **`client.close()` failure skips cleanup** (CodeRabbit) — fixed with `gather(..., return_exceptions=True)`.
- **`<worker scope>` → `<scope-key>`** (CodeRabbit) — fixed in `docs/configuration/agents.md`; the skill reference regenerated from it.
- **Empty passes retain stale rooms** (qodo) — not changed. The zero-room guard in `_reconcile_full_pass` predates this PR and is intentional: an empty enumeration cannot be told apart from a failed one, so directory-wide reconciliation is skipped and a warning is logged. This is documented under `threads export` in `docs/cli.md`, and the in-process path inherits exactly the CLI's behaviour.

basnijholt:
Second-round triage (Codex review), pushed in the latest commit:

- **Membership changes do not trigger synchronization** — fixed. Admitted `ROOM_LIFECYCLE` events now count as room activity, so a join, leave, or ban re-evaluates that room's exports and the membership check retracts what the leaver can no longer read. Coverage is the router's rooms plus each bot's own transitions (that is what the journal admits); rooms the router is not in still reconcile on the next message, reload, or restart.
- **Documented debounce behaviour** — doc fixed. The loop runs one pass about two seconds after the first event and batches everything that arrives in that window. Trailing-edge debounce would starve exports under continuous chat, so the behaviour stays and the sentence now describes it.
- **`tach.toml` formatting** — fixed.
- **PR description listed a deleted test file** — fixed.
- **Authoritatively empty full passes preserve stale rooms** — not changed. This is the pre-existing `_reconcile_full_pass` zero-room guard the CLI has documented under `threads export` for a while; it predates this PR and applies identically to both paths.
- **Crashed passes do not retry autonomously** — not changed, by design. Transient Matrix or journal failures are caught per room inside the pass and recorded as failures; an exception escaping the pass is a bug, and a self-rearming retry on a bug is the tight loop this PR deliberately removed. Pending work is kept and the next room activity, reload, or restart runs it.
- **Configuration topology changes strand old export trees** — not changed. Exports live inside the agent's own state tree (`agents/<name>/workspace/` or the private root), which MindRoom leaves in place when an agent is removed or re-scoped. Deleting only the export subtree from an otherwise abandoned state directory would be inconsistent partial cleanup; the ownership marker keeps such trees inert.

basnijholt:
Third-round triage (Codex), pushed:

- **Invited-room membership retraction** — fixed. Room activity now fires for every live `RoomMemberEvent` at admission entry, before the admission decision, so a bot alone in an invited room re-evaluates the room when another user leaves or is banned. The previous `ROOM_LIFECYCLE`-kind trigger is folded into this. A bot's *own* departure still cannot retract: once it has left it cannot verify membership, and the "unverifiable membership keeps exports" rule is pre-existing (#1504) and applies to the CLI too.
- **Docs overpromise cleanup** — fixed. The workspace-export section now states the zero-room reconciliation guard and points at the `threads export` guidance.
- **Conflict with `main`** — merged; the two generated skill-reference artifacts were regenerated from the merged docs.
- **Crash backoff** — still declined. Per-room Matrix and journal failures are already caught inside the pass and recorded as failures. An exception that escapes is a disk or YAML fault, which breaks more than exports; a reload or restart retries it. Bounded backoff would be a state machine dedicated to a bug path.

basnijholt:
Fourth-round triage (Codex): **recovered membership events** — fixed, the activity announcement now fires for every non-`HISTORY` `RoomMemberEvent`, test extended with a `RECOVERED` case. **Fleet check / smoke** — the plugin-fleet drift check stays red until `thread-export-plugin` is archived (documented in the description; archive PR mindroom-ai/thread-export-plugin#12 is ready), and merge waits for the smoke workflow.

PR #1957, opened 2026-09-04
Group external trigger deliveries into one thread with thread_key

basnijholt:
## Summary

- add optional `thread_key` to the external trigger payload
- on a `new_thread` trigger, the first delivery with a key posts the usual per-fire root and records the key; later deliveries with the same key continue that Matrix thread instead of opening another root
- thread key records live in the existing replay store, scoped per trigger, retained 7 days after the most recent delivery that used them
- `mindroom trigger send --thread-key ...`
- docs: new "Grouping Deliveries Into One Thread" section

## Why

Some upstream sources deliver one conversation as many events, for example every reply in a chat thread or every message in a direct-message conversation. With `new_thread`, each of those became its own Matrix thread: noisy for the room, and the agent answered each reply with a fresh session and no memory of the earlier ones. Grouping by a sender-chosen key keeps one Matrix thread and one agent session per upstream conversation, while unrelated events still get their own thread.

## Behaviour details

- The key is opaque to MindRoom; senders pick something that names the upstream conversation (ticket ID, upstream thread timestamp, DM channel ID).
- A continued delivery is a normal in-thread message (`m.relates_to` to the recorded root, `m.in_reply_to` the latest thread event) and does **not** carry the per-fire-root marker, so the agent treats it as a follow-up in an existing thread.
- Fixed `target_thread_id` triggers ignore `thread_key`.
- Replay files written before this change have no `threads` section and still load; malformed thread records fail closed like malformed nonce or event records.
- `event_id` idempotency is unchanged and independent of `thread_key`.

## Verification

- `tests/test_external_trigger_replay_store.py`: payload validation, per-scope lookup, expiry, retention refresh, legacy file compatibility, corrupt record rejection
- `tests/test_external_trigger_executor.py`: continued delivery relates to the recorded root and omits the per-fire-root marker
- `tests/api/test_external_triggers_api.py`: two keyed deliveries share one thread, a different key starts a new one, fixed-thread triggers ignore the key
- `tests/test_cli_trigger.py`: `--thread-key` in the signed body
- 100 tests across those suites plus `test_import_graph.py` pass; pre-commit (ruff, ty, vulture, docs skill references) passes on changed files

## Summary by Sourcery

Group external trigger deliveries by an optional sender-defined key so related events share one Matrix thread and agent context.

New Features:
- Add optional `thread_key` support for grouping related external trigger deliveries into a shared Matrix thread and agent session.
- Expose `--thread-key` through the trigger-send CLI and signed request payload.

Bug Fixes:
- Prevent failed or concurrent deliveries from creating duplicate threads, while preserving existing event-id idempotency behavior.

Enhancements:
- Persist per-trigger thread-key routing with scoped retention, atomic reservations, expiry, fencing, room safety, and compatibility with legacy replay stores.
- Continue keyed deliveries as ordinary in-thread messages without marking them as new per-fire roots.

Documentation:
- Document thread-key usage, retention, concurrency behavior, and fixed-thread semantics in the external trigger and CLI references.

Tests:
- Add coverage for payload validation, replay-store lifecycle and corruption handling, keyed delivery routing, failure recovery, fixed-thread behavior, executor metadata, and CLI signing.

## Summary by CodeRabbit

- **New Features**
  - Added optional thread keys to external trigger requests and the `trigger send` CLI command.
  - Deliveries with the same key now share a Matrix thread for `new_thread` triggers.
  - Thread associations are retained for seven days and scoped to the target room; fixed target threads are unaffected.
  - Invalid keys, including blank or overlong values, are rejected.
  - Concurrent first deliveries receive a retryable conflict response.
  - Failed initial deliveries release reservations for retry, while failed continuations retain their thread association.

- **Documentation**
  - Added usage guidance, retention details, scope limitations, concurrency behavior, and CLI examples.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1322,6 +1322,9 @@ Send one signed trigger request to MindRoom.
 │ *  --message                            TEXT   Trigger payload message. [required]     │
 │    --event-id                           TEXT   Optional idempotency event id.          │
 │    --title                              TEXT   Optional trigger title.                 │
+│    --thread-key                         TEXT   Optional key; deliveries sharing it     │
+│                                                land in one Matrix thread on new_thread │
+│                                                triggers.                               │
 │    --data-json                          TEXT   Optional JSON object for trigger data.  │
 │    --timeout                            FLOAT  HTTP request timeout in seconds.        │
 │                                                [default: 10.0]                         │
--- a/docs/external-triggers.md
+++ b/docs/external-triggers.md
@@ -217,6 +217,7 @@ After that, or after `rotate_trigger_key`, the next delivery with the key starts
 Opening a thread is atomic per key.
 While one delivery is posting the first root, a concurrent delivery with the same key receives `409 External trigger thread is being opened by another delivery`; retry it with a fresh signed request and the same `event_id`, and it will join the thread.
 A reservation whose first delivery fails is released immediately; a crashed delivery releases it after 5 minutes.
+Reservations are fenced: a delivery that outlives its reservation can neither bind its root over a newer reservation nor release it, so a slow first delivery cannot orphan the thread that replaced it.
 
 A follow-up that fails to send keeps the key bound to its thread, so a transient error never splits one conversation across two threads.
 
--- a/docs/external-triggers.md
+++ b/docs/external-triggers.md
@@ -131,6 +131,7 @@ For a non-admin caller, the target agent and room are the current agent and curr
 
 `target_thread_id` and `new_thread` are mutually exclusive.
 With `new_thread`, each delivery posts a room-level root and the responding agent answers in a new thread under it with a fresh session.
+A sender can group related deliveries on a `new_thread` trigger by passing the same `thread_key`; see [Grouping Deliveries Into One Thread](#grouping-deliveries-into-one-thread).
 
 For an administrator caller, `target_agent` and `target_room_id` can additionally target a different agent, team, or room.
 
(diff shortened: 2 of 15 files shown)


basnijholt:
Follow-up to the latest review round, now at c7150c54c:

1. **Reservation fencing** — done. `claim_thread_key` returns an opaque reservation token; `bind_thread_root` and `release_thread_key` only act on a pending record when the token matches, and a late binder is handed the root the key ended up bound to rather than overwriting it. Late-release and late-bind interleaving tests added (`test_late_failure_of_an_expired_reservation_cannot_release_the_new_owner`, `test_late_success_of_an_expired_reservation_cannot_bind_over_the_new_owner`). Lease renewal is not needed once stale owners cannot touch the record: an expired lease only means a newer delivery may take the key, and the old one can no longer interfere.

2. **Strict record validation** — done. `thread_event_id` and `reservation` must both be present; a pending record must carry a reservation and a bound record must not. The "missing `thread_event_id`" shape and three related ones are in the fail-closed parametrization.

3. **Single transaction for bind + mark-delivered** — not done, deliberately. The Matrix send can never be inside the store transaction, so a "sent but not fully recorded" window exists regardless of how the two writes after it are grouped. Merging them removes one of two windows at the cost of coupling event-id and thread-key records, and the residual failure shape (message delivered, event id stays in progress until its TTL, key reservation expires after 5 minutes) is the same shape `mark_event_delivered` already has for every trigger today. Happy to revisit if maintainers want post-send bookkeeping made best-effort as a whole, but that is a separate change from this feature.

PR #1987, opened 2026-09-10
Give journal replay one ordering owner per room

basnijholt:
A rotating journal scan could skip earlier messages while their room was busy or postpone a failed callback indefinitely as new events kept arriving. Repeated admissions could also bypass retry backoff.

Global scanning now discovers rooms, while one reserved lane per room reads fresh room-filtered journal pages and owns receipt order, progress, and continuation. Room retries use independent one-to-thirty-second cooldowns and continue even while global discovery is delayed. The journal remains authoritative for approval eligibility.

Retry handoffs carry their existing room identity and invalidate that room's current page synchronously before later callbacks can run. Response ownership retains the room ID; approval and ingress callbacks forward it through their existing wiring. Room progress remains owned while detached responses hold unsettled sources. Completion notifications rewind immediately; missed notifications are recovered by a rotating fallback sweep that repeats independently of discovery and yields between bounded batches. A room receipt marker survives page boundaries and puts repeated eligible work under the existing cooldown, bounding immediate and alternating downstream handoffs with constant memory per room. Shutdown cancels owned work without settling pending sources.

The worker shrinks from 558 to 513 lines. The change adds an indexed room filter to the existing replay query and removes failed-head ordering barriers. Changes in the bot lifecycle shell only forward room identity.

Validation: journal ingress/storage/crash recovery, durable ingestion, turn-controller, approval, coalescing, response, configuration, and shutdown suites passed, including SQLite and PostgreSQL. Regressions cover busy-room order, continuing admissions, blocked discovery, synchronous approval handoffs, concurrent drains, unreadable pages, backoff, and restart. Changed-file pre-commit checks passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when deferred events are released, retried, reclaimed, or recovered after an owner becomes unavailable.
  - Prevented duplicate event admission and retry loops during room recovery.
  - Preserved deferred work across shutdown and worker restarts.
  - Added room-aware retry cooldowns and more complete recovery scans.

- **Improvements**
  - Pending event queries can now be narrowed by room.
  - Approval and response handoffs retain room context for more accurate retry and recovery behavior.

### DIFF
--- a/docs/architecture/bot-runtime.md
+++ b/docs/architecture/bot-runtime.md
@@ -91,6 +91,7 @@ Successful progress past the failed receipt resets that room's retry delay.
 Room admission and retry history remain alive while downstream responses own unsettled sources, including after the room lane exits.
 Retry handoffs carry their room ID and invalidate that room's current page synchronously, without waiting for a source lookup or affecting another room's progress.
 Fallback liveness checks rotate through a bounded batch; the periodic sweep yields between batches so a large backlog does not multiply probes for every callback or delay the next sweep by a full period per batch.
+Once scheduled, the sweep repeats while deferred owners remain, independently of discovery or lane completion.
 Silent owner loss invalidates room admission when the fallback detects it, through the same room wake used by explicit completion notifications.
 Shutdown cancels lane and retry owners without settling unfinished work, which remains available for startup recovery.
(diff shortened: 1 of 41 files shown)


basnijholt:
Addressed the shutdown logging finding in 18eb93e2c: handler failures are logged before the stopped-state guard, while shutdown still creates no retry timer. The three CI failures expected immediate replay despite the new cooldown; those tests now verify that the running dispatcher settles the original event after its retry becomes eligible. All 434 journal-ingress, crash-recovery, and focused turn-controller tests passed locally, along with changed-file pre-commit hooks.

